PR #16, opened 2015-01-13
Fix missing newlines after template expansion

sorbits:
This is a repeat of https://github.com/bcosca/fatfree/commit/970c97580b84d07b856172683afd84ffb82a9b63 that someone since broke.


bcosca:
This should be merged against dev, not master.

bcosca:
Ugh sorry about that. I thought it was the fatfree repo.

PR #18, opened 2015-01-22
Update error handler to not error on `E_NOTICE`

sorbits:
This was previously submitted as bcosca/fatfree#646 — since then `E_NOTICE` has been exempt from default error reporting (bcosca/fatfree@29d77780c9954efc857030eb46836d7a56b54006) which makes this commit even more relevant.

Also exempted `E_USER_NOTICE` since that isn’t an error either.

### DIFF
--- a/base.php
+++ b/base.php
@@ -1844,7 +1844,7 @@ final class Base extends Prefab implements ArrayAccess {
 		@ini_set('magic_quotes_gpc',0);
 		@ini_set('register_globals',0);
 		// Intercept errors/exceptions; PHP5.3-compatible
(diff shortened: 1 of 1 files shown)


ikkez:
looks like this already happened in the current version:
https://github.com/bcosca/fatfree-core/blob/master/base.php#L1986
https://github.com/bcosca/fatfree-core/blob/master/base.php#L1996
nevertheless, thank you.

sorbits:
It happened in ee11337f59fc2f1e7e29d0a6e42ad6a4b61392f3 which was basically my commit with changed message/author and a reference to this PR, but for some reason @bcosca didn’t close the PR when he applied my fix.

PR #25, opened 2015-02-07
Move composer.js to composer.json. Remove self ref repository section

sam-at-github:
Minor tidy up in composer.json. Not sure why it was called "composer.js", but composer will not find it named this way, so moved it. Also removed redundant repositories section. Tested, works with:

    {
	    "name": "test",
	    "require": {
		    "bcosca/fatfree-core": "dev-master"
	    },
	    "repositories": [
		    {
			    "type": "vcs",
			    "url": "https://github.com/sam-at-github/fatfree-core"
		    }
	    ]
    }

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,12 +6,6 @@
 	"require": {
 		"php": ">=5.3.6"
 	},
-	"repositories": [
-		{
-			"type": "vcs",
-			"url": "https://github.com/bcosca/fatfree-core"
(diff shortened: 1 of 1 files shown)


sam-at-github:
Note once tags get pulled over from bcosca/fatfree as described in bcosca/fatfree#761 then you will be able to specify version properly from composer.

PR #28, opened 2015-02-12
Fix for CamelCase namespace and class files

binarygeotech:
Added fix for CamelCase namespace and class files on case-sensitive OS

Example:
To autoload this file 'Classes\Helpers\HelperClass.php' with class name 'helperClass', the autoloader can get the file 'HelperClass' using 'helperClass'

I've been facing this problem on several projects in which the only remedy is to rename my class files or class name.
With this fix, everything worked flawlessly.

### DIFF
--- a/base.php
+++ b/base.php
@@ -1719,6 +1719,7 @@ final class Base extends Prefab implements ArrayAccess {
 	protected function autoload($class) {
 		$class=$this->fixslashes(ltrim($class,'\\'));
 		$func=NULL;
+		$glob_class = glob($this->reg_case("{$class}.php"));
 		if (is_array($path=$this->hive['AUTOLOAD']) &&
 			isset($path[1]) && is_callable($path[1]))
 			list($path,$func)=$path;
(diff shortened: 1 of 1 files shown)


xfra35:
> several projects in which the only remedy is to rename my class files or class name

You seem to have missed the ability to customize the autoloader using a [custom function](http://fatfreeframework.com/routing-engine#Autoloadercasehandling), which in your case would have given:

```php
$f3->set('AUTOLOAD',array('classes/',function($class){
  return implode('/',array_map('ucfirst',explode('/',$class)));
}));
```

binarygeotech:
I tried using the autoloader and the custom function, but failed too. I raised an issue on the autoloader case sensitivity on case-sensitive OS some months ago, which I patched on my server. But I noticed there might be someone that will like to enjoy the flexibility of F3 with class files named without case-sensitivity in mind.

Here is what I did sometimes ago:

````
protected function autoload($class) {
		$class=$this->fixslashes(ltrim($class,'\\'));
		foreach ($this->split($this->hive['PLUGINS'].';'.
			$this->hive['AUTOLOAD']) as $auto)
{
$class_ar = explode('/',$class);
array_walk($class_ar, create_function('&$cls', '$cls = ucwords(strtolower($cls));'));
$u_class = implode('/',$class_ar);

			if (is_file($file=$auto.$class.'.php') ||
				is_file($file=$auto.strtolower($class).'.php') ||
				is_file($file=strtolower($auto.$class).'.php') ||
				is_file($file=$auto.$u_class.'.php'))
				return require($file);
}
	}
````

KOTRET:
your fix will array_walk all files every time any unknown class should be loaded. Furthermore there were already several issues addressing CaSE sENsitIviTy. Please customize the loading as suggested.

PR #30, opened 2015-02-24
Remove cookies from cached template header

ajjjjjjjj:
Related to thiss issue: https://github.com/bcosca/fatfree/issues/763


bcosca:
A more optimal solution has been committed.

PR #31, opened 2015-02-26
Fix composer.json - replace invalid key autoload with classmap.

sam-at-github:
"autoload" is not a valid key for the composer "autoload" field. Also lib/ only works for bcosca/fatfree. Have to use "." here. Refs bcosca/fatfree#775.

Composer autoloading now works and you do:

    require 'vendor/autoloader.php';
    $f3 = \Base::instance();

instead of:

    $f3  = require 'path/to/base.php';


KOTRET:
:+1: imho that fixes 7477134ed43d68bad9499b545ddaa0a3d00fab01 correctly

PR #32, opened 2015-02-26
Deprecate implicit construction

sam-at-github:
It's bad practice IMHO to do anything non declarative from an included file/module. The only reason this used to be done is to get the F3 autoloader to load. That is not needed given the changes to composer.json in #31. The f3 autoloader is still there underneath the composer autoloader. If your not using composer you now do:

    require 'path/to/Base.php';
    $f3 = \Base::instance();

Otherwise, with composer (#31) its:

    require 'vendor/autoloader.php';
    $f3 = \Base::instance();

### DIFF
--- a/base.php
+++ b/base.php
@@ -2929,4 +2929,3 @@ final class Registry {
 
 }
 
-return Base::instance();
--- a/composer.json
+++ b/composer.json
@@ -7,6 +7,6 @@
 		"php": ">=5.3.6"
 	},
 	"autoload": {
-		"autoload": ["lib/"]
+		"classmap": ["."]
 	}
 }


KOTRET:
exactly my thoughts, see https://github.com/bcosca/fatfree/issues/775#issuecomment-73699955 for that. This should be fixed.

sam-at-github:
> exactly my thoughts, see bcosca/fatfree#775 (comment) for that. This should be fixed.
Yeah I know I was following that one :).

ikkez:
so Base class is being required, even if not used? That just confirms that the composer autoloader is stupid.

KOTRET:
ikkez: no. if you declare it via `"autoload"` then yes, otherwise no. its correct in composer.

PR #34, opened 2015-03-05
Image: allow absolute paths

xfra35:
The Image constructor doesn't make any distinction between absolute and relative paths.

This PR intends to fix that. Additionally it raises an error if the file can't be found.

### DIFF
--- a/image.php
+++ b/image.php
@@ -570,7 +570,7 @@ class Image {
 			foreach ($file[0]=='/'?array(''):$fw->split($path?:$fw->get('UI').';./') as $dir)
 				if (is_file($dir.$file))
 					return $this->load($fw->read($dir.$file));
-			user_error(self::E_File);
+			user_error(self::E_File,E_USER_ERROR);
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


adaniello:
Thanks!

ikkez:
hm, can't we use the `$path` constructor argument for setting an absolute path already?

xfra35:
I thought about it too, changing:
```php
$fw->split($path?:$fw->get('UI').';./')
```
into
```php
$fw->split(isset($path)?$path:$fw->get('UI').';./')
```
But the `split()` method removes any *empty* element, which includes `''`.

Or did you mean splitting the filename in two? Like `new Image(basename($path),FALSE,dirname($path))` instead of `new Image($path)`? That would probably work, although a bit less user-friendly :P

ikkez:
Well, basically i thought about splitting the filename in two, yes. probably not that comfortable, but that's how i do it atm. I was not aware of the split function, that it removes the empty elements. Soo, i guess improving this thing is reasonable.

xfra35:
I made an alternative PR, which allows `$path=''`. I think this way, we're closer to the initial logic.

PR #35, opened 2015-03-06
Image + minify: allow absolute paths

xfra35:
Hi,
this is an alternative to https://github.com/bcosca/fatfree-core/pull/34.
Here we keep the existing logic, with the only difference that `$path` can be set to an empty string, thus allowing `$file(s)` to contain absolute paths.

### DIFF
--- a/base.php
+++ b/base.php
@@ -586,10 +586,11 @@ final class Base extends Prefab implements ArrayAccess {
 	*	Split comma-, semi-colon, or pipe-separated string
 	*	@return array
 	*	@param $str string
+	*	@param $noempty bool
 	**/
-	function split($str) {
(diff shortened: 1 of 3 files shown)


adaniello:
Well done!
Thanks

bcosca:
Wouldn't you be tempted to autoload classes from the filesystem root?

xfra35:
Doesn't look like a good practice to depend on PHP libraries located outside of the project folder.

PHP classes are critical components, images aren't. I think it makes sense, in some situations, to have a media storage area located outside of the web root.

bcosca:
In summary, the 2nd arg of split() should be used with extreme prejudice.

PR #38, opened 2015-03-09
fix config section detection

ikkez:
One of the unit tests is failing. There was an issue in the recent change made here: https://github.com/bcosca/fatfree-core/commit/538008b6e526b14ab78952441d9b9661e51ce9c5#commitcomment-10108618

the suggested regex fix and unit tests see: [Debuggex Demo](https://www.debuggex.com/r/ZLpKnKiFkcfbbPah)
Regular expression visualization


ikkez:
@bcosca does this fix your needs, or do you maybe have a smarter fix for this? Or is it okay like this?

PR #39, opened 2015-03-11
moved beforeerase trigger in erase()

dexwerx:
calls beforeerase trigger before any actions are taken.


ikkez:
closed due to bcosca/fatfree#801 and bcosca/fatfree#803
further discussion at bcosca/fatfree#797

PR #41, opened 2015-03-11
No implicit construction in base.php

sam-at-github:
One liner. rm `Base::instance()` on load construction. Bad practice. Causes issues. Had pushed this earlier, but accidently deleted PR branch.

### DIFF
--- a/base.php
+++ b/base.php
@@ -3008,5 +3008,3 @@ final class Registry {
 	}
 
 }
-
-return Base::instance();


jmversteeg:
I'm all for this since it is indeed really bad practice. It should be noted, though, that currently in the `README.md` the following example is given:
```php
$f3 = require('path/to/base.php');
```
(at https://github.com/bcosca/fatfree/tree/dev#hello-world-the-less-than-a-minute-fat-free-recipe)

Rayne:
This pull request breaks the backwards compatibility. I'm not expecting that @bcosca will drop this feature, so it would be a good idea to collect the issues.

I haven't experienced any issues although I am using composer and `Base::instance()` to get the instance.

KOTRET:
it will break backwards compatibility for those who are doing `$f3 = require('base.php');` and easy to fix. Since removing this line will fix some of the discussed issues it also cleans up bad codestyle. So imho there is nothing against this pr.

PR #45, opened 2015-04-10
Router fix: loose method search

xfra35:
This PR intends to fix [this](https://groups.google.com/forum/#!topic/f3-framework/hyhFDTA0XvY).

In a nutshell, when we define:
* `POST /foo`
* `GET /@token`

`GET /foo` should be valid (it returns 405 at the moment).

### DIFF
--- a/base.php
+++ b/base.php
@@ -1500,8 +1500,7 @@ final class Base extends Prefab implements ArrayAccess {
 				}
 				return $result;
 			}
-			$allowed=array_keys($route);
-			break;
+			$allowed=array_merge($allowed,array_keys($route));
 		}
 		if (!$allowed)
 			// URL doesn't match any route
--- a/base.php
+++ b/base.php
@@ -1507,7 +1507,7 @@ final class Base extends Prefab implements ArrayAccess {
 			$this->error(404);
 		elseif (PHP_SAPI!='cli') {
 			// Unhandled HTTP method
-			header('Allow: '.implode(',',$allowed));
+			header('Allow: '.implode(',',array_unique($allowed)));
 			if ($this->hive['VERB']!='OPTIONS')
 				$this->error(405);
 		}


ikkez:
if we are about to touch this, let's review https://github.com/bcosca/fatfree-core/pull/12 again.

bcosca:
Whatever you say guys. I got your backs covered :)

ikkez:
Well, it's a little change in behaviour, but I think it's a good move into the right direction and makes the route handling more intuitive. I must admit that I hit this issue serveral times too. Go for it @xfra35, push the button ;)
In addition: I'm not sure if it is necessary, but a system variable like `STRICT` could be used here to switch between a strict or loose route & verb matching. #12 should also be included... makes totally sense and works great in some tests.

xfra35:
I'm ready to push :D
Just to clarify, what would be the difference between loose and strict modes?

ikkez:
well, strict mode stops on the first match, and if no verb handler was found you get the 403. Loose mode would continue to match against all other defined routes until a handler was found, or it's an OPTIONS request... while writing this, something comes to my mind and I found an issue with this:

```php
$f3->route('GET /r/@r',function( Base $f3, $params) {
	echo 'R@'.$params['r'];
});
$f3->route('GET /r/*',function( Base $f3, $params) {
	echo 'R*'.$params[1];
});
$f3->route('GET /r-test',function( Base $f3, $params) {
	$f3->mock('OPTIONS /r/2');
	var_dump(headers_list());
});
```
The header now contains: `Allow: GET,GET`, which is obviously wrong.

xfra35:
Nice catch. I fixed it.

As for the strict mode: what would be the benefit of using it?

ikkez:
only benefit would be backward compatibility (you want that error), and a more strict route handling that you maybe need for APIs. But well, nevermind... I guess no one actually would really need this, and could also besolved with some route configs, but maybe we keep this in mind and add this if someone asks.

PR #46, opened 2015-04-10
View: Removed reserved variables $fw and $implicit

Rayne:
This pull request removes the reserved sandbox variable `$fw` and cleans the sandbox from default `$implicit` definitions.

Now it's possible to use the `$fw` variable in HIVEs and `$implicit` is no longer part of sandbox environments.

Remaining reserved variables: `$this` and `$hive`.

We could also remove `$hive` from the reserved variables by copying it to a temporary member variable.

### DIFF
--- a/base.php
+++ b/base.php
@@ -2463,9 +2463,9 @@ class View extends Prefab {
 			if (isset($hive['ALIASES']))
 				$hive['ALIASES']=$fw->build($hive['ALIASES']);
 		}
+		unset($fw, $implicit);
 		extract($hive);
 		unset($hive);
-		unset($fw);
 		ob_start();
 		require($this->view);
 		$this->level--;


xfra35:
I never knew I could access the framework with a simple `$fw`! I find it cool actually ^^
Maybe to rename it to `$f3`?
Or cleaner: make it a property of View so that we can access it with `$this->fw`.

Rayne:
You can't access it by `$fw` as `$fw` gets removed from the sandbox. But this also makes the variable `$fw` not usable for custom HIVEs.

I think the bigger problem is that the `$implicit` variable, which I introduced with the security patch for [Broken escaping for sub-Views with default (and custom) HIVE](https://github.com/bcosca/fatfree/issues/749), got leaked into the sandbox.

xfra35:
Ah yes true, I missed the previous `unset`. Then your PR makes sense. Let's wait for @bcosca's review.

ikkez:
> Maybe to rename it to $f3?

this would actually be nice. i think it is not necessary to make it a class property,
 so we could use `{{@f3}}` in templates directly.

Rayne:
Is it necessary to have access to the `Base` instance in a sandboxed `View` environment? What are the benefits? I prefer the clean sandbox (when defining a custom HIVE).

Adding the `Base` instance to the default HIVE is a piece of cake:

```php
$f3 = require('path/to/base.php'); // or $f3 = \Base::instance() with Composer
$f3->set('f3', $f3)
```

Rayne:
Please merge this request @bcosca and let us resume the discussion in a dedicated request or the Google Group.

PR #49, opened 2015-04-19
improved mapper & collection type hints

ikkez:
With this improved type hints, you can get autocompletion on collections (arrays that contain a specific type of objects), if your IDE supports the recommended way of from phpDoc http://phpdoc.org/docs/latest/references/phpdoc/types.html#arrays

screenshot-2015-04-19-18 51 50


Rayne:
:thumbsup:

adaniello:
:+1:

PR #50, opened 2015-04-20
configurable token filters

ikkez:
I reviewed [an old approach](https://github.com/ikkez/fatfree/commit/6ed71c53d8f0ddeb96ddc6a7bd71d9a519a9b734) of mine to make the template token filter / modifiers better configurable. Currently the only way to add new filters is to extend the template engine, which is basically great and really simple. Nevertheless, I had to struggle with that more than once yet, because I often use multiple template extentions and custom tags, where I always had to pass around my custom template instance to make own filters work in there.
So here comes `Preview->filter`, an easy way to add new token filters and to receive all already known filters, when called without parameter. The `View->alias` method becomes obsolete with this patch,.. it is now defined directly without a detour. This patch also makes it easier for 3rd party plugins to work along with the same global template instance from the prefab registry.

### DIFF
--- a/base.php
+++ b/base.php
@@ -2435,16 +2435,6 @@ class View extends Prefab {
 	}
 
 	/**
-	 *	build an url from alias name
-	 *	@return string
-	 *	@param $key string
-	 *	@param $arg string
-	 **/
-	function alias($key,$arg=null) {
-		return Base::instance()->alias($key,$arg);
-	}
-
-	/**
 	*	Create sandbox for template execution
 	*	@return string
 	*	@param $hive array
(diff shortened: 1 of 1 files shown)


xfra35:
Great :+1:
It was missing for too long :P

Can we keep `View->alias` though? There are situations where it's needed. For example:

* `<set target="{{$this->alias('foo')}}"/>` (filter not allowed here)
* `{~ $target=$this->alias('foo') ~}`
* etc..

adaniello:
:+1:

ikkez:
@xfra35 we could use the trick from #46, and rename the `$fw` to `$F3` and do not unset it in the sandbox. If so, you could simply write `<set target="{{$F3->alias('foo')}}"/>` aaaaand we could replace the `\Base::instance()->...` with `$F3->...` for the other filter definitions too. That would reduce recurring calls to the Registry and gives a little performance increase.

xfra35:
Yes why not. An alternative is to add a pointer to the framework from `View`: `$this->alias()` becomes `$this->f3->alias()`. This would be, I guess, an acceptable syntax and would leave the sandbox environment clean from predefined variables.

Rayne:
I think that providing `$f3` as a member variable is an acceptable solution.

If one still needs `$f3` as variable in the template/view's context, one could add it with `$f3->set('f3', $f3)` or as custom hive member.

daslicht:
Are there anywhere example how to use those?
Is it possible that extending the Template no longer works ?

PR #53, opened 2015-04-29
CORS support, bcosca/fatfree#731

ikkez:
This PR goes with https://github.com/bcosca/fatfree/issues/731
It follows the general [CORS flowchart](https://camo.githubusercontent.com/9ea48e5f02e4aaea34dc7f7584ee1bb0680e098e/687474703a2f2f7777772e68746d6c35726f636b732e636f6d2f7374617469632f696d616765732f636f72735f7365727665725f666c6f7763686172742e706e67) as well as some recommendations from https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS.

It can be controlled by the following settings:

* **CORS.headers**, string or array, allowed headers
* **CORS.origin**, string or false, the allowed origin host or wildcard
* **CORS.credentials**, bool, if cookies are allowed
* **CORS.expose**, string or array, which headers to expose to the client browser
* **CORS.ttl** int, caching time of the preflight OPTION request, see [addition info](http://stackoverflow.com/questions/23543719/cors-access-control-max-age-is-ignored) why a simple constant should be sufficient here.

The `CORS` defaults disallows everything, so it's just disabled as usual. A simple `CORS.origin = *` would enable basic cross-site request support.

### DIFF
--- a/base.php
+++ b/base.php
@@ -1519,7 +1519,7 @@ final class Base extends Prefab implements ArrayAccess {
 			// Unhandled HTTP method
 			header('Allow: '.implode(',',array_unique($allowed)));
 			if ($cors) {
-				header('Access-Control-Allow-Methods: '.implode(',',$allowed));
+				header('Access-Control-Allow-Methods: OPTIONS,'.implode(',',$allowed));
 				if ($cors['headers'])
 					header('Access-Control-Allow-Headers: '.(is_array($cors['headers'])?
 						implode(',',$cors['headers']):$cors['headers']));
--- a/base.php
+++ b/base.php
@@ -1394,6 +1394,13 @@ final class Base extends Prefab implements ArrayAccess {
 		$this->hive['ROUTES']=array_combine($keys,$vals);
 		// Convert to BASE-relative URL
 		$req=$this->rel($this->hive['URI']);
+		if ($cors=(isset($this->hive['HEADERS']['Origin']) &&
+			$this->hive['CORS']['origin'])) {
+			$cors=$this->hive['CORS'];
+			header('Access-Control-Allow-Origin: '.$cors['origin']);
+			header('Access-Control-Allow-Credentials: '.
+				($cors['credentials']?'true':'false'));
+		}
 		$allowed=array();
 		foreach ($this->hive['ROUTES'] as $pattern=>$routes) {
 			if (!$args=$this->mask($pattern,$req))
@@ -1424,6 +1431,9 @@ final class Base extends Prefab implements ArrayAccess {
 				// Save matching route
 				$this->hive['ALIAS']=$alias;
 				$this->hive['PATTERN']=$pattern;
+				if ($cors && $cors['expose'])
+					header('Access-Control-Expose-Headers: '.(is_array($cors['expose'])?
+						implode(',',$cors['expose']):$cors['expose']));
 				if (is_string($handler)) {
 					// Replace route pattern tokens in handler if any
 					$handler=preg_replace_callback('/@(\w+\b)/',
@@ -1508,6 +1518,14 @@ final class Base extends Prefab implements ArrayAccess {
 		elseif (PHP_SAPI!='cli') {
 			// Unhandled HTTP method
 			header('Allow: '.implode(',',array_unique($allowed)));
+			if ($cors) {
+				header('Access-Control-Allow-Methods: '.implode(',',$allowed));
+				if ($cors['headers'])
+					header('Access-Control-Allow-Headers: '.(is_array($cors['headers'])?
+						implode(',',$cors['headers']):$cors['headers']));
+				if ($cors['ttl']>0)
+					header('Access-Control-Max-Age: '.$cors['ttl']);
+			}
 			if ($this->hive['VERB']!='OPTIONS')
 				$this->error(405);
 		}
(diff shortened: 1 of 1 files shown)


adaniello:
:+1: 

Many thanks!

bcosca:
I know a lot of introspection went into this piece of code, and I really appreciate that you referred this matter to me considering that we're introducing another global variable. +1 to everyone. BTW, can we push for a mid-year release of 3.4.1?

laapsaap:
God, please let this be the end of the CORS nightmare I am having for weeks now. Thank you very much.

dabiddo:
Does anyone have an example on how to use it?, I'm writing a small API, and can't get the CORS to answer request, this is what I have tried.

function dump($f3){
        $f3->set('CORS.headers','Access-Control-Allow-Origin:*');
        $f3->set('CORS.origin','*');
        $f3->set('CORS.expose','Access-Control-Allow-Origin:*');

        echo "OK";
    }

If I do a var_dump of the CORS array, I can see my parameters clearly.

ikkez:
It's just `$f3->set('CORS.origin','*');`. 
This will set the `Access-Control-Allow-Origin: *` header at the appropriate time.
With CORS.expose you can tell, which extra/custom headers the client is able to see n the response, CORS.headers is there to tell the client which headers are allowed to send in the request. http://fatfreeframework.com/quick-reference#CORS

AndrewEQ:
Hey ikkez, I've tried ```$f3->set('CORS.origin','*');``` but couldn't get the header to be set?!

Had a look at base.php and found this:

```
		if ($cors=(isset($this->hive['HEADERS']['Origin']) &&
			$this->hive['CORS']['origin'])) {
			$cors=$this->hive['CORS'];
			header('Access-Control-Allow-Origin: '.$cors['origin']);
			header('Access-Control-Allow-Credentials: '.
				($cors['credentials']?'true':'false'));
		}
```

...so had to set both of these before it worked, is it suppose to be this way?

```
$f3->set('HEADERS.Origin','*');
$f3->set('CORS.origin','*');
```

ikkez:
>  is it suppose to be this way?

[yes](https://camo.githubusercontent.com/9ea48e5f02e4aaea34dc7f7584ee1bb0680e098e/687474703a2f2f7777772e68746d6c35726f636b732e636f6d2f7374617469632f696d616765732f636f72735f7365727665725f666c6f7763686172742e706e67). Usually your browser will send that Origin header on its own when you attempt to do a real cross-domain request. So you won't see that header being send when the request is from the same domain. Using jquery you can force that setting with the crossDomain option (which should be true for cross-domain requests by default):
```js
$.ajax({
	type: "GET",
	crossDomain: true,
	dataType: 'json',
	url: URL+'&'+data
})
```
Regarding your 2nd question: Since the whole CORS processing and detection of relevant routes and available methods is something that can only take place in the `run` method, the only suitable approach to have CORS enabled only for `/api/` routes, is IMO to have A.) a real directory `/api/` containing another `index.php` which runs F3 with other settings, or B.) use a [middleware router](https://github.com/ikkez/F3-Sugar/tree/master-v3/Middleware) or other custom checks on the current `$_SERVER['REQUEST_URI']` to set that setting or not.

AndrewEQ:
1) Ok, I see the protocol, great!
2) The docs aren't clear on when/where to set CORS.origin, had to look at the base to figure it out: http://fatfreeframework.com/quick-reference#CORS
A) Only great if you're making an api with sub domains
B) Most times you just want to define the CORS where you need it and not in the index.php

boussou:
Hi, I was surprised Fat Free was not handling CORS correctly.
That PR did not help to make my Cordova client work.
The proper anwsers that actually work are here: 

javascript - CORS with php headers - Stack Overflow
http://stackoverflow.com/questions/8719276/cors-with-php-headers

putting a call to  **cors()**  before anything solve it.
FYI

ikkez:
HI. The F3 router does the same as the things in that cors() function. I guess it's only a matter of configuration. The anwsers suggest to set the ORIGIN to the server host and not *. This is probably the correct answer.

boussou:
True.
when you set " * " it is rejected.

I was not able to understand how it works with the framework.
What's important is to reply to options with the $_SERVER['HTTP_ORIGIN'] for Access-Control-Allow-Origin.
The answer to OPTIONS request must not have a content-type, but using $f3->run() there exist a content type which end up in an error message in the browser.

kavin-90:
You can do HTTP ORIGIN in apache or nginx too.
You can enable XSS protection from webserver too

boussou:
Why are you saying this?
My remark was just to help F3 work out of the box.

kavin-90:
i'm helping F3 by saying that because F3 made double rules so XSS protection got disabled and no sniff also got disabled because it got invalid.

boussou:
Please elaborate because it is not very clear.

kavin-90:
webserver enabled XSS protected and no sniiff and origin so F3 rules will be invalid.

boussou:
what webserver? it is not apache apparently?

kavin-90:
Nginx

boussou:
can you share your nginx conf?

kavin-90:
Header set X-XSS-Protection "1; mode=block"

ikkez:
I guess it's not F3 which adds the additional header, but your nginx config doubles the value. The webserver actually sends the final response, so when you add a header in here, F3 is already done with it. If possible, add a conditional statement in your config, and only add it when the header is not already there. Or just leave that part to the framework.

kavin-90:
can you give me an example?

PR #62, opened 2015-05-30
Update template.php

binarygeotech:
In previous version, I was able to use the include tag this way without error;

```` <include href="{{@temp_base}}.{{@section}}.'header.html'" /> ````

after updating to the latest version, I received

``` 500 Internal Server Error ```, in the trace, I saw ```` {{@temp_base}}.{{@section}}.'header.html' not found ````.

it only accepts

```` <include href="{{@temp_base}}.{{@section}}.{{@header_file}}" /> ````
 
I checked through the file history and I noticed the change was added as `` Teak Regex `` on `` March 4, 2015 ``;
```` preg_match('/^\{\{(.+?)\}\}$/',$attrib['href']) ```` which restricts the ``` href ``` value to declared variables or strings only, mixture not allowed.

This little fix save me from the error by allowing mix value in the href attribute of the include tag.

### DIFF
--- a/template.php
+++ b/template.php
@@ -73,7 +73,7 @@ class Template extends Preview {
 			'<?php '.(isset($attrib['if'])?
 				('if ('.$this->token($attrib['if']).') '):'').
 				('echo $this->render('.
-					(preg_match('/^\{\{(.+?)\}\}$/',$attrib['href'])?
+					(preg_match('/^\{\{(.+?)\}\}/',$attrib['href'])?
 						$this->token($attrib['href']):
 						Base::instance()->stringify($attrib['href'])).','.
 					'$this->mime,'.$hive.'); ?>');

### CODE REVIEW

**Comment by ikkez** [comment]
okay, I see. The change was introduced here: https://github.com/bcosca/fatfree-core/commit/ebb1928fea6993c9d51e6d1b214993e76a2bdfe4
@bcosca what was the matter of that change? Can we revert that?

**Comment by ikkez** [comment]
I think this was made to better distinguish static and dynamic attributes. A soultion for your case would be:

```html
<include href="{{@temp_base.@section.'header.html}}'" />
```

binarygeotech:
Yeah, I tested it and it works with `` <include href="{{@temp_base.@section.'header.html'}}" /> ``, but what of those that doesn't know they can do it this way, the change was introduced without prior notification which broke the template on existing project.
This patch is just a removal of the strict restriction that it must end with `` }} ``, it should accommodate any usage.
These scenarios should all work without error;

`` <include href="{{@temp_base.@section.'header.html'}}" /> ``
`` <include href="{{@temp_base.@section}}.'header.html'" /> ``
`` <include href="{{@temp_base}}.{{@section}}.'header.html'}}" /> ``

ikkez:
Well usually you can put static strings into the `href`-attribute or a token to express a dynamic value. It is difficult to maintain the tag handlers in the template engine if you mix both types. Hence you got the possibility to concatenate values and variables in a token. Maybe the documentation lacks a bit of explaining it correctly, but even your samples are not following a strict syntax

`<include href="{{@temp_base.@section}}.'header.html'" />`
Why should this work? When a mix should be possible, it need to be
`<include href="{{@temp_base.@section}}header.html" />`. Otherwise this static sample `<include href="'tmpl/home/header.html'" />` with extra single quotes should make no difference, which is wrong.
But as i said, this doesn't work and make it more complicated to code a tag handler that works with all edge cases.

We have only have static attributes
`<include href="tmpl/home/header.html" />`
and a token value (php expression) for dynamic vars in your attributes
`<include href="{{@temp_base.@section.'header.html'}}" />`

Maybe this is more strict than before, but it forces you to code in a more clean way. Actually it was a bug that it has worked before ;)

binarygeotech:
All well said, the documentation needs to be updated for clear explanation on this.

PR #63, opened 2015-05-31
Preview/Template/View: added the `$this->f3` property discussed in #50

Rayne:
This patch adds the `View->f3` property discussed in https://github.com/bcosca/fatfree-core/pull/50.
It passes the `View` and `Template` unit tests and is a better alternative than escaping and cloning the whole `Base` object (repeatedly) by defining `Base::instance()->set('f3', Base::instance());` or `$customHive = ['f3' => Base::instance(), /* … */]`.
Also it's compatible with `PHP 5.3`, which [isn't able to escape and clone objects](http://fatfreeframework.com/base#recursive).
Please keep in mind that only accessible non-static properties are escaped with `PHP >= 5.4` and that `Base->$hive` is a private property.

As we aren't interested in escaping and cloning the `Base` instance to use its methods, we shouldn't escape and clone it.

I'm not satisfied by this patch and would recommend to use `Base::instance()` instead of `$this->f3`, but it could be useful for others:

* https://github.com/bcosca/fatfree-core/pull/50#issuecomment-94781039
* https://github.com/bcosca/fatfree-core/pull/50#issuecomment-94674124

Now it's your turn to decide the fate of this pull request. Have fun!

### DIFF
--- a/base.php
+++ b/base.php
@@ -2425,6 +2425,8 @@ class Cache extends Prefab {
 class View extends Prefab {
 
 	protected
+		//! Fat-Free Framework
+		$f3,
 		//! Template file
 		$view,
 		//! post-rendering handler
@@ -2432,13 +2434,17 @@ class View extends Prefab {
 		//! Nesting level
 		$level=0;
 
+	function __construct() {
+		$this->f3 = Base::instance();
+	}
+
 	/**
 	*	Encode characters to equivalent HTML entities
 	*	@return string
 	*	@param $arg mixed
 	**/
 	function esc($arg) {
-		$fw=Base::instance();
+		$fw=$this->f3;
 		return $fw->recursive($arg,
 			function($val) use($fw) {
 				return is_string($val)?$fw->encode($val):$val;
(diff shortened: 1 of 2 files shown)


ikkez:
Thanks for recalling this Rayne. Regarding the issue, I think there is no real enhancement here,.. for me it's excess. Putting the Base instance into the hive itself feels wrong. It's actually easy to access the Base in template by using the Base, like `<set target="{{\Base::instance()->alias('foo')}}"/>`. If you'd like to have a shortcut for this, one would think that the var `f3` or `F3` is used for the Base instance anyways, which takes me back to https://github.com/bcosca/fatfree-core/pull/50#issuecomment-94758518, and have it as always predefined var in the sandbox (that would also bypass the escaping thing). that's why I am not for a member variable here.

Rayne:
I discussed the patch with @ikkez and our consensus is, that `Base->alias()` should be used instead of the (now missing) shortcut `View->alias()`. The shortcut was never intended to be used like in the linked (no longer working) examples.

PR #65, opened 2015-05-31
Update Base::map() to take object as well as string handler.

sam-at-github:
Update Base::map() to take object as well as string handler. Akin to route() taking a callable.

### DIFF
--- a/base.php
+++ b/base.php
@@ -1300,20 +1300,28 @@ final class Base extends Prefab implements ArrayAccess {
 	*	Provide ReST interface by mapping HTTP verb to class method
 	*	@return NULL
 	*	@param $url string
-	*	@param $class string
+	*	@param $handler string|object
 	*	@param $ttl int
 	*	@param $kbps int
 	**/
-	function map($url,$class,$ttl=0,$kbps=0) {
+	function map($url,$handler,$ttl=0,$kbps=0) {
 		if (is_array($url)) {
 			foreach ($url as $item)
-				$this->map($item,$class,$ttl,$kbps);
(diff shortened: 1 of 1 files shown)


xfra35:
What would be a use case for this? Usually, a controller class is instantiated at routing time, not before.

BTW, your 2nd commit improperly instantiates *all* classes.

sam-at-github:
> What would be a use case for this?
Routing to an existing object.
> Usually, a controller class is instantiated at routing time, not before.
Well, I'm not sure about what is the usual way, but Base::route() can take a callable. This commit makes Base::map() consistent with Base::route().
My use case is just this - it was convenient for me to init and configure my handler object before routing. That way I can even pass constructor args.
> BTW, your 2nd commit improperly instantiates all classes.
I don't follow. @xfra35 please explain me.

xfra35:
> Base::route() can take a callable

Point taken. Implemented in https://github.com/bcosca/fatfree-core/commit/845d9a4b9f9deb48a7e26c05e2de5062927c545d.

Still, I'm curious to know a configuration in which instantiating the controller class before routing is useful. Could you share an example?

sam-at-github:
> Still, I'm curious to know a configuration in which instantiating the controller class before routing is useful. Could you share an example?

I have a JSON RPC class that wraps other model classses like this `new JsonRpcWrapper(new ModelClass())`. The JsonRpcWrapper takes care of mapping the return values, and errors into the correct JSON RPC. I could create a sub class of every model class, but that would be smelly.

xfra35:
I see now. Thanks for sharing. Well now, you can do it.

PR #67, opened 2015-06-10
Abort: flush output

xfra35:
With this change, buffered output is flushed instead of being discarded, thus allowing `echo` statements prior to `abort()`:

```php
$f3->route('GET /',function($f3) {
  echo 'Running in background';//<-- not ignored
  $f3->abort();
  //etc...
});
```

### DIFF
--- a/base.php
+++ b/base.php
@@ -1577,9 +1577,11 @@ final class Base extends Prefab implements ArrayAccess {
 	function abort() {
 		@session_start();
 		session_commit();
-		header('Content-Length: 0');
+		$out='';
 		while (ob_get_level())
-			ob_end_clean();
+			$out=ob_get_clean().$out;
+		header('Content-Length: '.strlen($out));
+		echo $out;
 		flush();
 		if (function_exists('fastcgi_finish_request'))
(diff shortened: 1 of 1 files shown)


Rayne:
Looks good and it should be safe to use `strlen`, as HTTP counts the octets/bytes for `Content-Length`. (I assume that we ignore `mbstring.func_overload`, like everyone does.)

xfra35:
@Rayne interesting comment about `mbstring.func_overload`. I'm pretty sure that it would break other parts of the framework as well.

[One way](http://php.net/manual/en/function.mb-strlen.php#77040) to take it into account would be to use `mb_strlen($out, '8bit')` instead. But let's ignore this issue for now, as long as nobody complains...

ikkez:
:+1: I just tested it and seems to work well.

ikkez:
@bcosca Is this good to go with your initial intention to this function? I think having an output flush here makes this feature much more valuable for other tasks too.

bcosca:
This is great. I haven't felt the need for a flush because I have no use case for any output; but this code addition is absolutely correct.

Issue #68, opened 2015-06-24
TZ auto detection


xfra35:
Hi,

[Currently](https://github.com/bcosca/fatfree-core/blob/master/base.php#L2137), TZ is autodetected using `ini_get('date.timezone')`.

I've just encountered the case on a server where this autodetection didn't work because `ini_get('date.timezone')` returned nothing. On the other hand, `date_default_timezone_get()` did return the correct timezone, so why don't we use `date_default_timezone_get`?

bcosca:
Good idea. PHP >5.4 has a different autodetection mechanism, so `ini_get('date.timezone')` may be outdated.

PR #69, opened 2015-06-26
TZ detection with date_default_timezone_get()

Rayne:
This PR implements https://github.com/bcosca/fatfree-core/issues/68. The `@` suppresses the warning which occurs when `PHP < 5.4` fails to guess a time zone.


ikkez:
:+1:

PR #72, opened 2015-07-14
Added support for {@token} syntax in route definition

nanawel:
It allows to handle routes like these (line 2-3):
```
GET /=app\controllers\IndexController->index
GET /account/@action=app\controllers\AccountController->{@action}Action
GET /account/@action/*=app\controllers\AccountController->{@action}Action
GET /account_operation/@action=app\controllers\Account\OperationController->@action
GET /account_operation/@action/*=app\controllers\Account\OperationController->@action
```
which enables you to add prefix or suffix to the tokens found in the URL in order to generate the final handler method (or class, etc.).
**I needed this more importantly to "hide" public methods of the controllers that I didn't want to be reached by URL.** (in my case, only methods ending in "Action" can be called)

If it's not necessary because of some other cool feature that I didn't know about, I'd be glad to learn :)

### DIFF
--- a/base.php
+++ b/base.php
@@ -1369,7 +1369,7 @@ final class Base extends Prefab implements ArrayAccess {
 			$url=$this->rel($this->hive['URI']);
 		$case=$this->hive['CASELESS']?'i':'';
 		preg_match('/^'.
-			preg_replace('/@(\w+\b)/','(?P<\1>[^\/\?]+)',
+			preg_replace('/((\\\{\\\{)?@(\w+\b)(?(2)\\\}\\\}))/','(?P<\3>[^\/\?]+)',
 			str_replace('\*','([^\?]+)',preg_quote($pattern,'/'))).
 				'\/?(?:\?.*)?$/'.$case.'um',$url,$args);
 		return $args;
@@ -1437,10 +1437,10 @@ final class Base extends Prefab implements ArrayAccess {
 						implode(',',$cors['expose']):$cors['expose']));
 				if (is_string($handler)) {
 					// Replace route pattern tokens in handler if any
-					$handler=preg_replace_callback('/({)?@(\w+\b)(?(1)})/',
+					$handler=preg_replace_callback('/({{)?@(\w+\b)(?(1)}})/',
 						function($id) use($args) {
-							$tokid=count($id)>2?2:1;
-							return isset($args[$id[$tokid]])?$args[$id[$tokid]]:$id[0];
+                            $pid=count($id)>2?2:1;
+							return isset($args[$id[$pid]])?$args[$id[$pid]]:$id[0];
 						},
 						$handler
 					);
--- a/base.php
+++ b/base.php
@@ -1439,8 +1439,8 @@ final class Base extends Prefab implements ArrayAccess {
 					// Replace route pattern tokens in handler if any
 					$handler=preg_replace_callback('/({{)?@(\w+\b)(?(1)}})/',
 						function($id) use($args) {
-                            $pid=count($id)>2?2:1;
-							return isset($args[$id[$pid]])?$args[$id[$pid]]:$id[0];
+							$tokid=count($id)>2?2:1;
+							return isset($args[$id[$tokid]])?$args[$id[$tokid]]:$id[0];
 						},
 						$handler
 					);
--- a/base.php
+++ b/base.php
@@ -1437,12 +1437,13 @@ final class Base extends Prefab implements ArrayAccess {
 						implode(',',$cors['expose']):$cors['expose']));
 				if (is_string($handler)) {
 					// Replace route pattern tokens in handler if any
-					$handler=preg_replace_callback('/@(\w+\b)/',
-						function($id) use($args) {
-							return isset($args[$id[1]])?$args[$id[1]]:$id[0];
-						},
-						$handler
-					);
+                    $handler=preg_replace_callback('/({)?@(\w+\b)(?(1)})/',
+                        function($id) use($args) {
+                            $tokid=count($id)>2?2:1;
+                            return isset($args[$id[$tokid]])?$args[$id[$tokid]]:$id[0];
+                        },
+                        $handler
+                    );
 					if (preg_match('/(.+)\h*(?:->|::)/',$handler,$match) &&
 						!class_exists($match[1]))
 						$this->error(404);
--- a/base.php
+++ b/base.php
@@ -1437,13 +1437,13 @@ final class Base extends Prefab implements ArrayAccess {
 						implode(',',$cors['expose']):$cors['expose']));
 				if (is_string($handler)) {
 					// Replace route pattern tokens in handler if any
-                    $handler=preg_replace_callback('/({)?@(\w+\b)(?(1)})/',
-                        function($id) use($args) {
-                            $tokid=count($id)>2?2:1;
-                            return isset($args[$id[$tokid]])?$args[$id[$tokid]]:$id[0];
-                        },
-                        $handler
-                    );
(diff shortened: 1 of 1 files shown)


nanawel:
Hi,
Is there any issue with this patch? Should I modify anything?
If not, when can we hope to see it pushed in the main branch?

Thanks in advance

Rayne:
I like the idea.  :thumbsup:

I haven't tested your patch, but it looks like it doesn't support multiple `@arguments`.

Examples:

```
GET /@module/@action=app\controllers\{@module}Controller->{@action}Action
GET /@partA/@partB=app\controllers\PartController->{@partA}And{@partB}Action
```

The first example could be useful whereas the second one looks like a bad designed route or controller.

nanawel:
Your second example also works with this modification.

Defining the following route
```
GET /@partA/@partB=app\controllers\PartController->{@partA}And{@partB}Action
```

and then accessing the following path from the browser
http://www.myhost.com/foo/bar

gives the following controller call:
```
app\controllers\PartController->fooAndbarAction
```
(tested on my branch)

Isn't it the desired behavior?

Rayne:
That's great! Let's see what @ikkez and @bcosca think.

ikkez:
Well, even if this is only a small modification, I'm not fully convinced that this adds any clarity to the router. It looks like a valuable addition, thanks for the effort, but I would have assumed to be able to use the `{ }` encapsulation on both sides of the route. But that indeed leads to the similar issue https://github.com/bcosca/fatfree/issues/832. 
IMO the issue you have, can already be solved pretty easy in other ways. 
You could use the [PREMAP](http://fatfreeframework.com/quick-reference#PREMAP) framework variable to have a common name prefix for your controller actions. Or you just separate your controller classes into ActionController and another one, or work with public and protect functions.

nanawel:
Using PREMAP seems to be another way to resolve my issue. Using public and protected methods is not an option though, as several public methods are not actions in my controllers.

An improved syntax with braces would have been cool too, but I understand the implications you're describing.

nanawel:
Hmm it seems that I cannot use PREMAP after all. This only applies to ReST requests mapping which I don't use. So my conclusion is that I still need this braces syntax.

I'm going to study the \Base::mask() method in order to support it on both sides of the route definition.

xfra35:
An alternative would be to extend the `PREMAP` behaviour to all routes (not only those mapped to HTTP verbs). It depends if we need this braces syntax only for method prefixing or if it can be useful for something else.

nanawel:
You're right. But in this case would it be better to use `PREMAP` for both cases or add another framework variable? Adding a new use to an already defined variable could lead to regressions in existing applications.

xfra35:
Well, considering the fact that `PREMAP` has only been here [for 2 months](https://github.com/bcosca/fatfree-core/blob/master/CHANGELOG), I guess that breaking backwards compatibility on such a short period could be an option.

bcosca:
I'm open to this change, but if we're going with this approach I suggest we use `{{ @token }}` to make it consistent with the rest of the framework.

nanawel:
@bcosca What change are you referring to? Using `PREMAP` also for routing or adding support for braces syntax?

For the first case I would suggest to add a `PREROUTE` variable instead. For the second I agree that the syntax should be consistent with what already exists elsewhere in the framework.

bcosca:
I think PREROUTE and PREMAP are mutually exclusive, since the latter is intended for ReST methods.

nanawel:
SInce `PREROUTE` seems like a good addition but is different from being able to handle route definitions the way the `{{@token}}` syntax does, I still think this latter option should be considered too.

I've modified the `mask()` method so that it now handles this syntax on the left side of the route. The syntax now also uses double curly braces instead of one.

I've tested it successfully with the following definition, which - I think - uses all the possible cases:
```
GET /@module/{{@controller}}/{{@action}}/*={{@module}}\@controller->{{@action}}Action
```
and it gives when routing **http://www.myhost.com/foo/bar/baz/params**:
```
foo\bar->bazAction
```
which seems right to me.

What do you think?

ikkez:
The problem I see with using `{{ @token }}` syntax in routes is, that double curly braces will conflict with the `$allow` parameter in Base->config, which is used to resolve those tokens for dynamic variables in the whole .ini file. That could become funny.

bcosca:
I don't think the braces are appropriate for the lval.

nanawel:
How can I check that my modification doesn't conflict with the `$allow` parameter?
Could you provide me with an example of file I could use on my branch please?

ikkez:
No that wont fit. I think the only way it could work is your very first intention
 `GET /@module/@action=app\controllers\{@module}Controller->{@action}Action`

nanawel:
So... what should I do to see this feature be integrated in the main branch, whatever its form?

xfra35:
Well... what are our options here?

1) Introduce a single brace token. That would lead to 3 types of token (no brace, 1-brace, 2-brace):
```ini
[routes]
GET /{{@subdir}}/@module/@action = {{@controllers_dir}}\@module->{@action}Action
```

2) Extend the `PREMAP` behaviour to all routes. That would allow method prefixing but not class prefixing. NB: I don't think that `PREMAP` and `PREROUTE` are mutually exclusive.

3) Introduce a new hook, triggered in `call()` just before the function call, which would allow to manipulate the target class and method.

4) Keep the double brace syntax but prevent template tokens from being resolved inside `[routes]` and `[maps]` sections.

Any other idea?

Considering those 4 options, I'd go for the first one, as it's easy to understand and doesn't break backward compatibility. For the sake of consistency, I'd also allow the single braces on the left part.

ikkez:
I think we don't really need that modification at all. It can be solved with proper class inheritance and using public / protected methods.

nanawel:
I'd go for the first one too. For the left part of the route definition it's easy to modify my code to support the single braces instead of the double.

xfra35:
Well, let's wait for @bcosca to take a decision on this.

I also don't need that modification myself but I can understand that it may be interesting in some situations. For example, if you want to make a *visible* distinction between the methods that directly handle routes and those which are simply helpers. Like: 
```php
public function action_read($f3) {// route handler (identified by the prefix `action_`)
  $f3->set('records',$this->fetchRecords());
}
protected function fetchRecords() {// helper
  //..
}
```
Isn't it the reason why you need that feature @nanawel ?

nanawel:
This is exactly the reason why I needed this. I have controllers with 3 or more levels of inheritance, with several helpers methods in the parent classes. Some of these methods are also callable from outside because I also have some "main" classes that are not controllers (which is not really easy with F3 since it is very "controller-centered"). However, the behavior of these methods depends on the context: the exact active controller for the current request. So each subclass is allowed to override helper methods if needed.

Not sure this is clear, but again, this cannot be considered a fool request.

ikkez:
If we go for 1.) then left side braces would be useful for such things:
```
GET /image/{@width}x{@height}/@action = Controller\Image->{@action}Action
```
If this is easy to do, I'm okay with it. But any further magic that is expensive should excatly be checked, as the routing mechanism is a critical point in terms of performance.

nanawel:
Actually I personally think the braces syntax should be the only one supported since it's the only way to allow freedom in naming and at the same time avoid confused interpretations. But since there must be backward compatibility, we'll need to support both.

xfra35:
@ikkez good example :+1:

@nanawel can you please make a new clean PR with both left and right single braces so that @bcosca  can easily see the changes we agreed on and hopefully press the merge button? :)

nanawel:
@xfra35 Sure, I'll do that asap.

bcosca:
It's a go :)

bcosca:
Awaiting new PR for single brace syntax.

nanawel:
New PR #79 pending.
Thank you all.

PR #74, opened 2015-07-28
don't import external files for minification

ikkez:
this wont work anyway, but also drops the whole import statement, i.e. for external font/css includes


ikkez:
one example that currently fails is
```css
@import url("https://fonts.googleapis.com/css?family=Open+Sans:300italic,400italic,700italic,400,300,700");
```

Issue #75, opened 2015-08-13
template parsing single tag elements


ikkez:
I noticed that the [Template->parse](http://fatfreeframework.com/template#parse) method could act a little bit weird when you have a [registered tag handler](http://fatfreeframework.com/template#extend) and have a html element in your markup that only has a starting tag, without being properly self-closed. 

This **works**:
```html
<set key="val" />
```
but this **breaks** the markup:
```html
<set key="val">
```

This looks trivial and easy to solve by adding the `/ >` tag ending, but sometimes I don't recognize it immediately and this construct is eating up some of your subsequent html markup that just get lost. I have a tag handler on `<input>` fields, which are common, single tag elements that are HTML5 valid without being self-closed. 
I have no problem with adding the closing slash, but some skeleton frontend frameworks like h5bp, bootstrap, etc. are shipped with the simplified syntax `<input type="text" name="search">` and it isn't really reliable for plugin developers to force the rewrite of templates. 
I haven't found a fix for the [parse](https://github.com/bcosca/fatfree-core/blob/master/template.php#L270) method yet, but I can imagine that a bit juggling with `$stack` and `$depth` could do the trick.

bcosca:
Let me see if I can help you with that.

xfra35:
The dummy `$empty` argument looks misleading (as if an extra information is sent to the callback).

Why not adding an extra `$empty` property to the Template class, holding the list of empty tags? This would require to initialize it with `array('set','include')` and to add a third argument to the `extend()` method.

bcosca:
That sounds logical. I'll see what I can do later.

ikkez:
Ah okay, you try to skip increasing the `$depth` by using a flag from the handlers methods parameter. That's a neat hack, very creative ;) but the ReflectionFunction isn't properly working here, when you define a custom function as string like `Foo::renderATag`. And maybe there is a better way, not to use a reflection at all, as it might slow down the rendering a lot, when you handle a lot of tags. Let's have a look at the tag `$stack` again.. maybe we can find the needle.

bcosca:
@xfra35 is right. This approach is indeed more optimal as it does away with all those dummy arguments and reflection.

ikkez:
I couldn't sleep well last night because of this :D It took me a while to understand how this smart template parser was build, really very great work bong. 
But I thought that the root issue here must be somewhere in the way how the parser works. The way the node tree is currently build: 
[...]

PR #77, opened 2015-08-22
improved Template parsing

ikkez:
#### It's up to **10x** faster now :laughing: 
The solution for #75 worked so far, but I worked on the parse method to make it needless to set flags for empty tags. As describes in here https://github.com/bcosca/fatfree-core/issues/75#issuecomment-131055889 it now uses a reverse way for tree depths building on closing tags, instead of increasing the tree growth on starting tags.
I also found a way to add chunked-parsing of the template string, which increases the total parsing speed as well :D for a little benchmark I looped the parsing of my [fooforms template](https://github.com/ikkez/F3-Sugar/blob/master-v3/FooForms/ui/templates/fooforms.html) which contains dozens of different handled tag elements.
```php
$file = $f3->read('ui/templates/fooforms.html');
$start = microtime(TRUE);
for ($i = 0,$max = 10;$i<$max;$i++)
	\Template::instance()->parse($file);
$end = microtime(TRUE);
echo "time: ".round(1e3*($end-$start),2)."ms,  '.
  'memory: ".round(memory_get_usage(TRUE)/1e3,1)."\n";
```
results [current](https://github.com/bcosca/fatfree-core/blob/c71683bb6b24d2fcce41f2c9f0b2b9c6e84a965e/template.php) solution:
```
time: 1666,97ms,  memory: 1572,9kb
```
with [5-chunked](https://github.com/bcosca/fatfree-core/commit/d9d5de81f7e701dfc7fe40fd273fc0ed55ddf872) parsing:
```
time: 369,09ms,  memory: 1572,9kb
```
with [adaptive-chunked](https://github.com/bcosca/fatfree-core/commit/a6cdae2b101aaf0701ca098b40b7151b9c9d23c7) parsing;
```
time: 158,79ms,  memory: 1572,9kb
```

### DIFF
--- a/template.php
+++ b/template.php
@@ -269,24 +269,24 @@ class Template extends Preview {
 	**/
 	function parse($text) {
 		// Build tree structure
-		for ($ptr=0,$len=strlen($text),$tree=array(),$tmp='';$ptr<$len;)
-			if (preg_match('/^<(\/?)(?:F3:)?'.
+		for ($ptr=0,$w=5,$len=strlen($text),$tree=array(),$tmp='';$ptr<$len;)
+			if (preg_match('/^(.{0,'.$w.'}?)<(\/?)(?:F3:)?'.
 				'('.$this->tags.')\b((?:\h+[\w-]+'.
 				'(?:\h*=\h*(?:"(?:.*?)"|\'(?:.*?)\'))?|'.
 				'\h*\{\{.+?\}\})*)\h*(\/?)>/is',
 				substr($text,$ptr),$match)) {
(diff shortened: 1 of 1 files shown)


Rayne:
The reversed tree depth parser concept is a great idea and the performance gain is a great achievement. Also all tests are passing. :+1:

ikkez:
Which test is failing? It's working here (latest commit): http://f3.ikkez.de/template

Rayne:
The second test was with the wrong branch and the first with the right. Sorry for the confusion. There are no failing tests.

PR #85, opened 2015-09-25
allow anonymous function for template filter

ikkez:
Regarding bcosca/fatfree#872, I thought it would be good to allow anonymous filters.

```php
\Template::instance()->filter('rev',function($val){
	return strrev($val);
});
```

Anonymous functions are already possible for *Template->extend* too, so this could make both usages more even. I'm just still indecisive with the implementation, cause you would need to clear your template-cache when you change your filter definition from string to callback. On the other hand we probably want to keep it simple and not outsource too much logic into the template. Thoughts?

### DIFF
--- a/base.php
+++ b/base.php
@@ -2569,7 +2569,10 @@ class Preview extends View {
 			$str,$parts)) {
 			$str=trim($parts[1]);
 			foreach (Base::instance()->split($parts[2]) as $func)
-				$str=$this->filter($func).'('.$str.')';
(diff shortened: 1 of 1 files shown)


Rayne:
:+1: for the PR and a caveat in the documentation.

Right now I can't think of a flexible compiled solution without using `eval()` (or something equivalent bad) for non-anonymous functions. Especially not when working with "non-standard" filters like `X::instance()->doY()->doZ`. The PR looks like a good compromise.

PR #86, opened 2015-09-29
Updated Preview->filter() documentation

Rayne:


Rayne:
Sorry, I opened this PR too early. It depends on https://github.com/bcosca/fatfree-core/pull/85 which isn't accepted yet.

PR #87, opened 2015-10-06
Bug fix: Cursor->load not always mapping to first record

xfra35:
This bug arises when using multiple `load()` on the same mapper:
```php
$mapper->load('');//10 records
$mapper->last();//ptr=9
echo $mapper->dry()?'ERR':'OK';// output: OK
$mapper->load('tag=6');//1 record found but ptr=9 doesn't map
echo $mapper->dry()?'ERR':'OK';// output: ERR
```

The reason of this bug is that `$this->skip(0)` is called before the pointer is reset.


ikkez:
:+1:

PR #93, opened 2015-10-25
Set default parameters for smtp class

ptdev:
Allows using the smtp class with local mailserver by just using "new \SMTP()" instead of having to define all parameters.


ikkez:
Yeah why not. I think it's a good addition.

Issue #89, opened 2015-10-12
DB-managed sessions autostart


xfra35:
At the moment, Cache & DB-managed sessions are automatically started:
```php
$f3=require('lib/base.php');
new Session();// calls session_start()
$f3->run();
```
If you don't want that behaviour, you have to replace a single call to `new Session()` with a multitude of calls every time you need a session. This makes it difficult to move an existing codebase from one session handler to another.

Couldn't we remove that automatic start? Looking at the code, it looks like whatever comes after `register_shutdown_function('session_commit')` in the [constructor](https://github.com/bcosca/fatfree-core/blob/cec6b90cdb92a41f862ecc413860f476ad19d714/session.php#L164) could be moved into the `open()` method.

bcosca:
This looks reasonable. Can you submit a PR for all session handlers?

xfra35:
That was actually trickier than it looked. Cf. #94

PR #94, opened 2015-11-04
DB-managed sessions autostart

xfra35:
This is a followup of #89. The original idea was to move the `session_start` out of the class constructor, in order not to autostart DB-managed sessions.

This change actually required more refactoring that I initially thought.

Here are a few points related to [PHP sessions](http://php.net/manual/en/function.session-set-save-handler.php) to consider:

* `session_start` triggers consecutively the `open` and `read` methods
* `session_write_close` triggers consecutively the `write` and `close` methods
* `session_write_close` is automatically called during the shutdown sequence unless it has been called specifically inside the script
* `session_commit` is an alias for `session_write_close`
* `session_destroy` triggers consecutively the `destroy` and `close` methods
* `session_regenerate_id` doesn't trigger any session class method
* `session_regenerate_id(TRUE)` triggers the `destroy` method only
* nothing prevents a script from calling `session_start` and `session_write_close` several times

Based on this behaviour, this PR suggests to refactor the session class as follows:
* CSRF token, agent string and client IP are computed/fetched in the constructor and written to db by the `write` method.
* suspicious sessions are detected inside `read`.
* the mapper `load()` method is called once, inside `read`. There's no reason to call it from `write`, as `write` should always come after `read`.
* `$sid` contains the session id **only** if the session has been started. So `(bool)$session->sid()` should be equivalent to `session_status()===PHP_SESSION_ACTIVE`, with the difference that it's working in PHP 5.3.

Additionally, I've included two changes in the class behavior which are not directly related to the autostart issue:
* as the `destroy` method is expected to clear the session data, and nothing more, I removed the cookie deletion from it. Since the cookie deletion is a framework feature, I think it better belongs to `$f3->clear('SESSION')`.
* the detection of suspicious sessions include invalid posted anti-CSRF tokens

Those 2 additional changes may not satisfy everybody as they may break backwards compatibility (slight change in `destroy` + anti-CSRF token expected to be transmitted over `POST.csrf`), so I'll wait for general agreement before patching the other session handlers.

### DIFF
--- a/base.php
+++ b/base.php
@@ -423,8 +423,7 @@ final class Base extends Prefab implements ArrayAccess {
 				// End session
 				session_unset();
 				session_destroy();
-				unset($_COOKIE[session_name()]);
-				header_remove('Set-Cookie');
+				$this->clear('COOKIE.'.session_name());
 			}
 			$this->sync('SESSION');
 		}
--- a/db/sql/session.php
+++ b/db/sql/session.php
@@ -27,7 +27,15 @@ class Session extends Mapper {
 
 	protected
 		//! Session ID
-		$sid;
+		$sid,
+		//! Anti-CSRF token
+		$_csrf,
+		//! User agent
+		$_agent,
+		//! IP,
+		$_ip,
+		//! Suspect callback
+		$onsuspect;
 
 	/**
 	*	Open session
@@ -44,6 +52,8 @@ class Session extends Mapper {
 	*	@return TRUE
 	**/
 	function close() {
+		$this->reset();
+		$this->sid=NULL;
 		return TRUE;
 	}
 
@@ -53,8 +63,20 @@ class Session extends Mapper {
 	*	@param $id string
 	**/
 	function read($id) {
-		if ($id!=$this->sid)
-			$this->load(array('session_id=?',$this->sid=$id));
+		$this->load(array('session_id=?',$this->sid=$id));
+		$fw=\Base::instance();
+		if (($ip=$this->get('ip')) && $ip!=$this->_ip ||
+			($agent=$this->get('agent')) && $agent!=$this->_agent ||
+			($csrf=$fw->get('POST.csrf')) && $csrf!=$this->get('csrf')) {
+			if (!isset($this->onsuspect) ||
+				FALSE===$fw->call($this->onsuspect,array($this,$id))) {
+				//NB: `session_destroy` can't be called at that stage (`session_start` not completed)
+				$this->destroy($id);
+				$this->close();
+				$fw->clear('COOKIE.'.session_name());
+				$fw->error(403);
+			}
+		}
 		return $this->dry()?FALSE:$this->get('data');
 	}
 
@@ -65,19 +87,11 @@ class Session extends Mapper {
 	*	@param $data string
 	**/
 	function write($id,$data) {
-		$fw=\Base::instance();
-		$sent=headers_sent();
-		$headers=$fw->get('HEADERS');
-		if ($id!=$this->sid)
-			$this->load(array('session_id=?',$this->sid=$id));
-		$csrf=$fw->hash($fw->get('ROOT').$fw->get('BASE')).'.'.
-			$fw->hash(mt_rand());
 		$this->set('session_id',$id);
 		$this->set('data',$data);
-		$this->set('csrf',$sent?$this->csrf():$csrf);
-		$this->set('ip',$fw->get('IP'));
-		$this->set('agent',
-			isset($headers['User-Agent'])?$headers['User-Agent']:'');
+		$this->set('csrf',$this->_csrf);
+		$this->set('ip',$this->_ip);
+		$this->set('agent',$this->_agent);
 		$this->set('stamp',time());
 		$this->save();
 		return TRUE;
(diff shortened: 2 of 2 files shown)


Rayne:
It seems natural to submit the CSRF token as `POST.csrf`, still I am not sure if it is a good idea to hard-code it. 

The PR fixes the bug https://github.com/bcosca/fatfree/issues/878 for `DB\SQL\Session`, but not the other `Session` backends. We should also patch the other ones when merging the PR.

ikkez:
First impression is great. I'm not sure about the `POST.csrf` though, but I think it'll make it much more convenient to validate forms. I suggest to change `$csrf=$fw->get('POST.csrf')` to `$fw->exists('POST.csrf',$csrf)`, otherwise it may fail in paranoid error reporting mode. And maybe an additional check for `$f3->VERB == 'POST'` could help in edge cases.
Why you removed the `if ($id!=$this->sid)` part from the read method? Isn't it there to ensure that it's not loaded from db multiple times on a session_start?

xfra35:
I'm not sure either about the `POST.csrf`. On one hand, I like the idea of the framework automatically checking the token for us, which would make it very simple to use. On the other hand, it lacks the ability to customize the field name and it doesn't work with other verbs (PUT, DELETE, etc..).

My first thought was that, with the refactored code, a simple `$f3->get('POST.csrf')==$session->csrf` would be enough to check the token value on the app side (remember: `$session->csrf` is now replaced in `write()`, which occurs normally at the end of script). But on the other side, it implies that the app code should directly refer to the session handler, which doesn't fill right. We should be able to add/remove/switch session handlers without modifying the application code.

> Why you removed the `if ($id!=$this->sid)` part from the read method?
This is because the suggested refactoring implies that `load()` is called only once, when session starts. This follows the internal mechanism of PHP sessions.

ikkez:
>  it implies that the app code should directly refer to the session handler

well we could place the csrf token in the session itself upon write. That way it could be available with a simple `$f3->get('POST.CSRF')==$f3->get('SESSION.CSRF')`. This is maybe more cross-compatible. We should name the key it in uppercase letters, which follows the rules for framework variables and should be fine. If it's not implemented in the user forms, it doesn't get in your way. 

> This follows the internal mechanism of PHP sessions.

yes you're right. I just checked it myself with multiple random session starts... your tweaks seem to work great. :+1:

bcosca:
+1

xfra35:
Storing the token inside `SESSION.csrf` is a good idea. We could even leave the choice for the key, in the same way as it is performed in `Image->captcha`. So the signature would become:
```php
function __construct(\DB\SQL $db,$table='sessions',$force=TRUE,$onsuspect=NULL,$key=NULL)
```
What do you think?

ikkez:
I like it. Go for it. @Rayne @KOTRET ?

Rayne:
Looks good, DO IT! :muscle:

@xfra35 Am I right, that your proposal disables the verification of `$key` (e.g. `$key = "SESSION.csrf"`)  when `$key = null`? ([Documentation](http://fatfreeframework.com/plug-ins#captcha-images) and [Source](https://github.com/bcosca/fatfree-core/blob/8985e65099ce4f6fa472f1a1c8a0a0028e5aa5d6/image.php#L440-L441) of `Image->captcha()` as reference). That behaviour and the auto-updating of `$key` upon write could be really convenient.

xfra35:
@Rayne, you wrote:
> Am I right, that your proposal disables the verification of $key (e.g. $key = "SESSION.csrf") when $key = null?

Well not exactly. The proposal was to remove the automatic checking and to simply have `$key` enable/disable the saving of the token:
* `$key=NULL` => no saving
* `$key='SESSION.csrf'` => save token to `SESSION.csrf`

But a new problem arises here: in order to save the token to `SESSION.csrf` from within the session class, we need:
* either to autostart the session in the constructor, which was precisely what this PR was against
* or to manipulate the serialized data inside `write()`, which doesn't sound like a good idea

ikkez:
Can't you just do this in the read method?

xfra35:
No because the read method occurs just before `$_SESSION` gets populated. So even if we set `SESSION.csrf` at that stage, it will be replaced immediately by the old value.

I think we can't make the token automatically persistent. We'll be back with the existing behaviour, except that the new token could be stored in a global variable, which allows using session handlers transparently.

```php
// index.php
new DB\SQL\Session($db,'sessions',TRUE,NULL,'CSRF');
if ($f3->VERB=='POST' && $f3->get('POST.csrf')!=$f3->get('SESSION.csrf')) {
  // CSRF detected
}
$f3->copy('CSRF','SESSION.csrf');
```

ikkez:
yes you're right.

xfra35:
Ok done here: https://github.com/bcosca/fatfree-core/commit/e808c55966b7f15977e8a493087610cfa706bfbc

Automatic CSRF token verification has been removed, so we need to persist the token value ourselves.

NB: I've left the existing `csrf()` method for backwards compatibility. So its:
* either `$f3->copy('CSRF','SESSION.csrf')`
* or `$f3->set('SESSION.csrf',$session->csrf())`

xfra35:
NB: other session backends have been patched accordingly.

PR #96, opened 2015-11-09
New: ttl attribute in <include>

xfra35:
Usage:
```html
<include href="somefile.html" ttl="86400"/>
```

### DIFF
--- a/template.php
+++ b/template.php
@@ -69,6 +69,7 @@ class Template extends Preview {
 									\Base::instance()->stringify($pair[2]));
 						},$pairs)).')+get_defined_vars()':
 					'get_defined_vars()';
+		$ttl=isset($attrib['ttl'])?(int)$attrib['ttl']:0;
 		return
 			'<?php '.(isset($attrib['if'])?
 				('if ('.$this->token($attrib['if']).') '):'').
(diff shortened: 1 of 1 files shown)


ikkez:
:+1:

KOTRET:
:+1:

Rayne:
The PR doesn't work for me as it doesn't clear the cached result after `5` seconds. Am I using it wrong?

[Source with generated files](https://gist.github.com/Rayne/031112f6300c0970f96b)

I also noticed that `Cache` and `Preview` are handling `$ttl` with a significant difference:

**Cache->set()**

> If $ttl is 0 then the entry is saved for an infinite time. Otherwise, the specified time, in seconds, is used as TTL.

**Preview->render()** (which doesn't use the `$ttl` argument of `Cache`)

> if ($ttl) $cache->set($hash,$data);

KOTRET:
yes, the call to Cache at [base.php#2683](https://github.com/bcosca/fatfree-core/blob/master/base.php#L2683) must be
```php
$cache->set($hash,$data, $ttl);
```

xfra35:
Well spotted guys!

PR #101, opened 2015-11-30
Performance improvement: mapper->select() should return a fixed size array

cordje:
After patching {mapper.php, cursor.php, base.php} in order to use SplFixedArray rather than array(), there's a ~20% improvement in memory and cpu speed in query heavy pages. 

At first the performance gain appeared more than that, but that's because base.php was ignoring the SplFixedArray in the recursive function for cleaning, i.e., for XSS avoidance. [All items in f3's hive are duplicated in memory in order to apply the cleaning functions!]

I'm fairly confident that all uses of select()'s output (i.e., the SplFixedArray) have been patched to be able to operate on the SplFixedArray (since native array functions don't work). 

If the idea isn't accepted, then at least it served as an experiment to see what advantage there would be. Further optimisation work should be focused on the way that f3 cleans variables before they're printed out by templates (e.g., by instead wrapping all calls to hive variables rather than cleaning all hive variables).

### DIFF
--- a/base.php
+++ b/base.php
@@ -759,6 +759,12 @@ final class Base extends Prefab implements ArrayAccess {
 			$stack=array();
 		switch (gettype($arg)) {
 			case 'object':
+				if($arg instanceof SplFixedArray){
+					$copy=new \SplFixedArray(count($arg));
+					foreach ($arg as $key=>$val)
+						$copy[$key]=$this->recursive($val,$func, array_merge($stack,array($arg)));
+					return $copy;
+				}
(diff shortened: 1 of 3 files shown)


Rayne:
That's an interesting idea to use a fixed size array instead of a dynamic one which wastes `Θ(n)` memory. I am not sure if this idea gets accepted, but here are some ideas for your PR:

* Use `fatfree-core`'s code style (e.g. tabs instead of spaces)
* Consistent indentation
* Also patch the other databases

cordje:
Thanks for the feedback Rayne, please see the latest commit.

PR #104, opened 2015-12-01
Performance improvement: web->send() should call ob_end_clean() & readfile()

cordje:
In my own benchmarks, these two changes reduced time from 1.5s to 0.03s (median over 100 req's of a 40MB file).
See http://www.garfieldtech.com/blog/readfile-memory for an explanation of why readfile is faster.

### DIFF
--- a/web.php
+++ b/web.php
@@ -130,6 +130,7 @@ class Web extends Prefab {
 			return FALSE;
 		$size=filesize($file);
 		if (PHP_SAPI!='cli') {
+			if(ob_get_level()){ob_end_clean();}
 			header('Content-Type: '.($mime?:$this->mime($file)));
 			if ($force)
 				header('Content-Disposition: attachment; '.
@@ -138,22 +139,24 @@ class Web extends Prefab {
 			header('Content-Length: '.$size);
 			header('X-Powered-By: '.Base::instance()->get('PACKAGE'));
 		}
-		$ctr=0;
-		$handle=fopen($file,'rb');
-		$start=microtime(TRUE);
-		while (!feof($handle) &&
-			($info=stream_get_meta_data($handle)) &&
-			!$info['timed_out'] && !connection_aborted()) {
-			if ($kbps) {
-				// Throttle output
-				$ctr++;
-				if ($ctr/$kbps>$elapsed=microtime(TRUE)-$start)
-					usleep(1e6*($ctr/$kbps-$elapsed));
-			}
-			// Send 1KiB and reset timer
-			echo fread($handle,1024);
+		if(!$kbps){
+		    readfile($file);
+		}else{
+		    $ctr=0;
+		    $handle=fopen($file,'rb');
+		    $start=microtime(TRUE);
+		    while (!feof($handle) &&
+			    ($info=stream_get_meta_data($handle)) &&
+			    !$info['timed_out'] && !connection_aborted()) {
+				    // Throttle output
+				    $ctr++;
+				    if ($ctr/$kbps>$elapsed=microtime(TRUE)-$start)
+					    usleep(1e6*($ctr/$kbps-$elapsed));
+			    // Send 1KiB and reset timer
+			    echo fread($handle,1024);
+		    }
+		    fclose($handle);
 		}
-		fclose($handle);
 		return $size;
 	}
 
--- a/web.php
+++ b/web.php
@@ -2,7 +2,7 @@
 
 /*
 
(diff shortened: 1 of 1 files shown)


KOTRET:
while digging in the sourcecode i found that both functions use buffered streams:
fread : https://github.com/php/php-src/blob/master/ext/standard/file.c#L1815
readfile: https://github.com/php/php-src/blob/master/ext/standard/file.c#L1410 and then 
https://github.com/php/php-src/blob/master/main/streams/streams.c#L1410

the Difference is buffer size ([4k](https://github.com/php/php-src/blob/master/main/php.h#L243) by default in readfile) and method overhead. So what buffersize to choose? In test i noticed a good performancegain up to 64k buffersize, after that the throughput slightly dropped. Guess its safe to use 8k as buffersize. So for me the fix would be something like this:
```php
		$ctr=0;
		$handle=fopen($file,'rb');
		$bufsize=1024 * $kbps ? 1 : 8;
		$start=microtime(TRUE);
		while (!feof($handle) &&
			($info=stream_get_meta_data($handle)) &&
			!$info['timed_out'] && !connection_aborted()) {
			if ($kbps) {
				// Throttle output
				$ctr++;
				if ($ctr/$kbps>$elapsed=microtime(TRUE)-$start)
					usleep(1e6*($ctr/$kbps-$elapsed));
			}
			// Send 1KiB and reset timer
			echo fread($handle,$bufsize);
		}
```

cordje:
Thanks for the feedback and the links to the PHP source :)

RE Comment 1: Do you not agree that it is better to do the chunk looping in C, rather than re-implement it in PHP?
Your proposal still does it in PHP, and in my benchmark is still slower than readfile().

RE Comment 2: Can you explain why disabling the output buffer is a bad sideeffect (or a sideeffect of what)? As far as I understand, output buffering is not required in this case since we don't want to modify or control the output (the file's contents) or the headers. 

Given that PHP's manual for readfile instructs to "....ensure that output buffering is off with ob_get_level()", I don't believe that the designers believed that it would cause problems either. Could you help me understand why I shouldn't do this using a test case showing the bad behaviour?

KOTRET:
> Do you not agree that it is better to do the chunk looping in C, rather than re-implement it in PHP?

Yea, it would be better to pass that job to php instead of reimplementing this (guess independent of the buffersize the overhead would still be noticeable). Idk why I've altered the code this way - probably abberation :grimacing:  
So in case `$kbps` is empty, we should just `readfile($file);` and `return $size;`

> Can you explain why disabling the output buffer is a bad sideeffect (or a sideeffect of what)?

I thought about it again because i was not sure if there are cases where you do not want to disable the buffer... If not on cli, the headers are sent anyway, so there can't be content echoed before and its not possible to send several files concatenated one by one. But what on cli? Leave it enabled and do a readfile can still cause an oom. To remove this inconsistency its best to close all buffers in every case. 
Notice that you can stack outputbuffers, so this code will close only the inner buffer, but not additional surrounding buffers (use `while` instead of `if`).
```php
	/**
	*	Transmit file to HTTP client; Return file size if successful,
	*	FALSE otherwise
	*	@return int|FALSE
	*	@param $file string
	*	@param $mime string
	*	@param $kbps int
	*	@param $force bool
	**/
	function send($file,$mime=NULL,$kbps=0,$force=TRUE) {
		if (!is_file($file))
			return FALSE;
		$size=filesize($file);
		if (PHP_SAPI!='cli') {
			header('Content-Type: '.($mime?:$this->mime($file)));
			if ($force)
				header('Content-Disposition: attachment; '.
					'filename="'.basename($file).'"');
			header('Accept-Ranges: bytes');
			header('Content-Length: '.$size);
			header('X-Powered-By: '.Base::instance()->get('PACKAGE'));
		}
		while (ob_get_level())
			ob_end_clean();
		if (!$kbps) {
			readfile($file);
			return $size;
		}
		$ctr=0;
		$handle=fopen($file,'rb');
		$start=microtime(TRUE);
		while (!feof($handle) &&
			($info=stream_get_meta_data($handle)) &&
			!$info['timed_out'] && !connection_aborted()) {
			// Throttle output
			$ctr++;
			if ($ctr/$kbps>$elapsed=microtime(TRUE)-$start)
				usleep(1e6*($ctr/$kbps-$elapsed));
			// Send 1KiB and reset timer
			echo fread($handle,1024);
		}
		fclose($handle);
		return $size;
	}
```

cordje:
Thanks - agreed

Issue #110, opened 2016-01-05
Namespace F3 - become PSR-4 Compatible.


sam-at-github:
Hi.

F3 should have a top-level namespace AKA vendor namespace, in order to be PSR-4 compatible. Something like `F3\`, or `Fatfree\` or what owner wants. Yes you can use PSR-4 autoloaders and composer autoloaders with F3, but the top level namespace should belong to the local application not a dependency. F3 is stepping on shoes (mine at least). For example, I have a local only class `View`, but that name is taken, along with a bunch more by F3 so I have to change it.

This has come up before as bcosca/fatfree#741, but I wanted to re-raise it here. I will re-factor code base if there is consensus, but not bothering if it won't be accepted.

Note on BWC; It will break, but I'm not saying merge into master just yet. Upgrade is a simple matter of prepending `namespace F3` to existing PHP files.

xfra35:
I agree :+1:

I also faced name collisions a few times. Nothing that couldn't be worked around, but it would certainly be cleaner to have all the framework stuff stored in one dedicated place.

That could come in next major release.

Rayne:
I also agree but I think that it's too late for `v3` because of the backward compatibility breaking changes. I hope that `v4` will have a vendor namespace. :+1:

sam-at-github:
Just completed, looks good. Used vendor namespace "F3" as guess that would be the preference  given its succinct and all.

eazuka:
this is highly needed. i just faced same issues as stated above where F3 has already taken names and it started having conflicts with some of my classes. Though simple renaming of my classes fixed issues but that shouldn't be the case, Fatfree is still a dependency so it should act like one.

bcosca:
Given that F3 is the framework you rely on, it should be the app that should adapt - not the other way around. Would jQuery give up the $?

sam-at-github:
> Would jQuery give up the $ because you want it?

False analogy. Yes, Javascripts are expected to respect namespaces.

nanawel:
> Given that F3 is the framework you rely on, it should be the app that should adapt - not the other way around. Would jQuery give up the $ because you want it? F3 gives you a lot of freedom - but not bordering on anarchy.

I strongly disagree with that. Every class in PHP must use namespaces nowadays. Only native classes provided by the SPL do not need so, and that's how you know they are core classes. A framework is still a library, even if it's the core of an application.
[...]

PR #112, opened 2016-01-10
Update fatfree-core to use the top level vendor namespace F3. Refs #110

sam-at-github:
Update fatfree core to use the top level vendor namespace `F3`. Refs #110. Required following refactoring most of which was automatable:

  * Add `namespace F3` to all top level files programatically, and amend the namespace of files that already had a namespace to include prefix `F3`.
  * Replace all relative refs to non F3 classes in the to level with absolute refs.
  * Amend all absolute refs to F3 class in non top level classes.
  * Minor touch up to autoloader - use require_once() not require().
  * Fix some implicit refs to non prefixed namespaces in Preview::token().
  * The sample app and test at /bcosca/fatfree will have to be updated. Already done sucessfully. Will push shortly.

### DIFF
--- a/audit.php
+++ b/audit.php
@@ -20,6 +20,8 @@
 
 */
 
+namespace F3;
+
 //! Data validator
 class Audit extends Prefab {
 
--- a/auth.php
+++ b/auth.php
@@ -20,6 +20,8 @@
 
 */
 
+namespace F3;
+
 
 //! Authorization/authentication plug-in
 class Auth {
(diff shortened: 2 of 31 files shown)


mstaack:
@bcosca any infos about this PR  and F3 PSR4 compability?

bcosca:
No go until v4 comes around due to backward compatibility. I suggest we open up a different branch for this.

sam-at-github:
:+1: All tests passed, but a branch would be great to get this change some real life testing. I would start using that upstream branch if you created one.

PR #115, opened 2016-02-04
ensure special Content-Type and Content_Length headers are properly handled

ekhaled:
The way Fat Free populates the `HEADERS` in `hive` is great,
however, `Content-Length` and `Content-Type` headers are not prefixed with `HTTP_` in the `$_SERVER` array. Nevertheless they are headers.

These need to be handled specially.

I'm trying to build a REST api using Fat Free, and POST/PUT requests can send different kinds of data, so I have to check the content type and see whether it's `application/x-www-form-urlencoded` or `application/json` or `multipart/form-data` and handle the payload accordingly.
Right now I'm having to patch the `HEADERS` in the application during bootstrap.


ikkez:
fixed

PR #116, opened 2016-02-08
Add password_hash support to auth.php

max-weller:
With the current implementation of auth.php, it is not possible to use proper password hashing because you need the `pw` from the database to check the salted hash.

This pull request implements a new option `[ 'pwhash' => true ]`. If this is set, passwords are not compared directly but the recommended [`password_verify`](http://php.net/password_verify) function of php is used.

### DIFF
--- a/auth.php
+++ b/auth.php
@@ -38,29 +38,41 @@ class Auth {
 		$args;
 
 	/**
-	*	Jig storage handler
-	*	@return bool
-	*	@param $id string
-	*	@param $pw string
+	 * Password compare function with optional hashing
+	 * @return bool
+	 * @param $db_pw string
+	 * @param $try_pw string
+	 **/
+	protected function _check_pw($db_pw, $try_pw) {
+		if (isset($this->args['pwhash']) && $this->args['pwhash'] === true)
+			return password_verify($try_pw, $db_pw);
+		else
+			return $db_pw === $try_pw;
+	}
+
+	/**
+	 *	Jig storage handler
+	 *	@return bool
+	 *	@param $id string
+	 *	@param $pw string
 	*	@param $realm string
 	**/
 	protected function _jig($id,$pw,$realm) {
-		return (bool)
-			call_user_func_array(
+		$loaded = (bool) call_user_func_array(
 				[$this->mapper,'load'],
 				[
 					array_merge(
 						[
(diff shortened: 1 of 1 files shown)


max-weller:
Unit Test: https://gist.github.com/max-weller/b52db38824d29f9937b2

max-weller:
Unit Test Results:
```
% rm -f _test.sqlite; php f3_auth_password_hash_test.php
Using test db: /home/maxweller/Projekte/f3-core/_test.sqlite
Test user created successfully                          Pass
Correct login was accepted                              Pass
Wrong username was rejected                             Pass
Wrong password was rejected                             Pass
```

ikkez:
It would probably be better to add a way for a custom password comparison check instead of adding just **one** option. *password_verify* is not the only option someone could need. So instead of add a limited check pw function, reuse the `$args` array from the construct and add a custom anonymous function here that's been used to checking passwords, maybe like:

```php
$auth = new Auth($sqlMapper,array('custom' => function($user,$hash) {
  return \Bcrypt::instance()->verify($user->pw,$hash);
}));
```

As this could be difficult with the current implementation, another way is to remove the Jig, Mongo, and SQL auth types completely, and only use the auth plugin for 3rd party authentication services (LDAP, OAuth, OpenConnect, etc.). Using this for your own SQL datasets with an unknown password mechanism might cause just trouble and it was a pretty often discussed and misunderstood topic in the community

max-weller:
Hi,
I just updated my branch so it can be merged again. I'd really like it if f3 would finally support encrypted passwords, not making users of the framework [Plaintext Offenders](http://plaintextoffenders.com/about/) by default. 
The password_hash/password_verify methods are the officially recommended methods to do this, so I think a customizable password check is not that important. 

Kind regards,
Max

ikkez:
You can add a custom compare method now:

```php
$success = $auth->login($id,$password,NULL,function($pw,$hash) {
	return password_verify($pw,$hash);
});
```

KOTRET:
As creating the $auth-Instance is bound to one mapper at a time, the function should be passed into the constructor instead into the login-method.

ikkez:
alright, I've changed it now.

damjes:
Please also update documentation, because current doesn't say a thing about possibility of using i.e. password_verify.

PR #119, opened 2016-02-18
Proposal for route wildcards to match empty strings

xfra35:
I recently discovered that wildcards used in routes don't match empty strings. So:

* `/foo*` matches `/foo/` but not `/foo`
* `/*` matches `/foo` but not `/`

This looks a bit counter-intuitive as it goes against the usual meaning of the `*` wildcard used in regexes or shells.

Hence this proposal to have wildcards match empty strings as well. This way, a route like `GET /*` could really catch *everything*.

NB: the tweak in `array_multisort` is here for `/foo` to have higher precedence order than `/foo*`.

### DIFF
--- a/base.php
+++ b/base.php
@@ -1384,7 +1384,7 @@ final class Base extends Prefab implements ArrayAccess {
 		$case=$this->hive['CASELESS']?'i':'';
 		preg_match('/^'.
 			preg_replace('/((\\\{)?@(\w+\b)(?(2)\\\}))/','(?P<\3>[^\/\?]+)',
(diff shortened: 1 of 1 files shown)


ikkez:
:+1: basically a good idea, but since the masking has changed, I'd like to test it with some more routes first, to see if the matching is not effected by this and the routes are probably resolved with their parameters. i.e.:
```
/foo/*/@lastArg
/foo/*/@preLast/@last
/foo*/@foo
/*/@foo
/*/*
```

ikkez:
merged

kavin-90:
Thanks

PR #123, opened 2016-03-01
Add ability to specify a name when sending a file via the Web plugin

Mindiell:
I added this to the Web->send() method in order to be able to rename a file downloaded.
Not sure about the name in the last position (because you have to give all other args) but as this, nothing is broken with all code ;o)


Rayne:
Good addition. Too bad that the position of `$name` has to be counterintuitive (as long as we don't want to break backwards compatibility for minor changes).

* Add the new parameter to the docblock of `send()`

* Squashed Fat-Free Framework style:

  ```
  'filename="'.($name!==''?$name:basename($file)).'"');
  ```

  The `!==` supports the highly unlikely file name `string "0"`.

PR #124, opened 2016-03-01
Add ability to specify a name when sending a file via the Web plugin

Mindiell:
A better PR is always a good thing :o)


Rayne:
Looks good to me. :+1:

PR #125, opened 2016-03-07
Proposal: silent operator to skip startup error check

xfra35:
Here's a proposal to fix bcosca/fatfree#851.

With this change, the startup error check could be disabled by setting the error reporting level to 0.

```php
error_reporting(0);
$f3=require('lib/base.php');
```

one-liner:
```php
@$f3=require('lib/base.php');
// or
$f3=@require('lib/base.php');
```

I haven't checked with Composer, but unless it tweaks itself the error reporting, it should work.

### DIFF
--- a/base.php
+++ b/base.php
@@ -2019,7 +2019,7 @@ final class Base extends Prefab implements ArrayAccess {
 		@ini_set('magic_quotes_gpc',0);
 		@ini_set('register_globals',0);
 		// Intercept errors/exceptions; PHP5.3-compatible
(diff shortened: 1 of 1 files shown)


ikkez:
great idea :+1:

Rayne:
Simple and natural solution based upon plain `PHP`. :+1:

bcosca:
That' s the F3 spirit! :+1:

xfra35:
Justed tested. Works fine with Composer:
```php
require('vendor/autoload.php');
@$f3=Base::instance();
```

PR #131, opened 2016-04-13
Adding flush while sending big files

antoniovazquezaraujo:
Hello. Thanks for the work!
I think I have found a little fail: with big files, (60 MB) send does not send nothing.
Adding this simple code, now is working.
Cheers from Spain.
Antonio.

### DIFF
--- a/web.php
+++ b/web.php
@@ -152,6 +152,11 @@ class Web extends Prefab {
 			}
 			// Send 1KiB and reset timer
 			echo fread($handle,1024);
+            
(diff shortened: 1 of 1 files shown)


ikkez:
the unit test isn't working anymore since this fix. it screws up the response page since the output buffer is flushed https://github.com/bcosca/fatfree/blob/dev/app/web.php#L28-L30

xfra35:
I've added a flag to control that behaviour.

PR #137, opened 2016-06-16
Changed some php <5.4 calls to array() with 5.4 array notation

ptdev:
A previous commit had made these changes but apparently some array() calls were missed. This should take care of that.

### DIFF
--- a/auth.php
+++ b/auth.php
@@ -74,12 +74,12 @@ class Auth {
 	protected function _mongo($id,$pw,$realm) {
 		return (bool)
 			$this->mapper->load(
-				array(
+				[
 					$this->args['id']=>$id,
 					$this->args['pw']=>$pw
-				)+
+				]+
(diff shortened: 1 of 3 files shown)


auth.php
```diff
@@ -74,12 +74,12 @@ protected function _jig($id,$pw,$realm) {
...
 			$this->mapper->load(
-				array(
+				[
 					$this->args['id']=>$id,
 					$this->args['pw']=>$pw
-				)+
+				]+
 				(isset($this->args['realm'])?
-					array($this->args['realm']=>$realm):array())
+					[$this->args['realm']=>$realm):[]]
```
ikkez: does not look correct. I think it should be
`[$this->args['realm']=>$realm]:[])`

ptdev:
@ikkez :+1: you're absolutely right. Latest commit should fix it.

PR #138, opened 2016-06-21
Fixed `SQL->log()` documentation

Rayne:
The method `log()` clears the log instead of disabling it.

**Example**

```php
$db = new DB\SQL('sqlite::memory:');

$db->exec('SELECT 1');
var_dump($db->log());

$db->exec('SELECT 2');
var_dump($db->log(false));

$db->exec('SELECT 3');
var_dump($db->log());
```

**Result**

```
string(17) "(0.1ms) SELECT 1
"
NULL
string(17) "(0.0ms) SELECT 3
"
```

### DIFF
--- a/db/sql.php
+++ b/db/sql.php
@@ -263,7 +263,7 @@ class SQL {
 	}
 
 	/**
-	*	Return SQL profiler results (or disable logging)
+	*	Return SQL profiler results (or clear the log)
 	*	@param $flag bool
 	*	@return string
 	**/


xfra35:
Fixed in https://github.com/bcosca/fatfree-core/commit/65d814a3a9aa81c340975cb65c00db0b1644e6b2.

Rayne:
I feared that your patch will be the accepted solution as it overrules the `$log` argument of `exec()`. It's a simple and convenient solution, but also a horrible one. Too bad that it can't be fixed without breaking the backwards compatibility of the `SQL` class.

xfra35:
I also find it strange. What do you think of this proposal:
* add a new public `$db->logging` property (default: `TRUE`)
* switch the default value of the `$log` argument of `exec()` to `NULL`, which indicates to stick to `$db->logging`
* keep `log(FALSE)` for BC, as a shorthand for `$db->logging=FALSE`, but mark it as deprecated

Rayne:
Your proposal is a clever workaround. Is it necessary for `$db->logging` to be public? :+1:

xfra35:
How would you modify it otherwise?

Issue #140, opened 2016-06-22
Separate configuration for cache and sessions


nanawel:
Currently the sessions depend on the cache configuration. In my opinion there should be a way to configure each of them independently.
For example, you may want to purge cache without deleting the sessions.
Also, you may want to store sessions and cache on different Redis databases, or even different Redis instances (one with persistance on disk, one without).

What do you think?

nanawel:
Created PR #141

PR #141, opened 2016-06-23
Sessions custom cache

nanawel:
Following the issue #140 I created earlier, here's an implementation of how this could be achieved while keeping full compatibility with the existing code.

Example of code:
```
$cache = Cache::instance();    // Regular cache (depends on CACHE config variable)
$sessionCacheInstance = new \Cache('folder=var/sessions/');    // Custom cache for sessions only
$session = new \Session(null, null, $sessionCacheInstance);

\Base::instance()->set('SESSION.myvar', $myValue);   // Data will be stored in var/sessions/ no matter what CACHE config variable says
```

Any feedback welcome

### DIFF
--- a/session.php
+++ b/session.php
@@ -33,7 +33,9 @@ class Session {
 		//! IP,
 		$_ip,
 		//! Suspect callback
-		$onsuspect;
+		$onsuspect,
+        //! Cache instance
+        $_cache;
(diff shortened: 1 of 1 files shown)


xfra35:
Very smart :+1:

PR #144, opened 2016-07-06
PSR-3 compliant Log class

nanawel:
I've adapted the Log class to comply with PSR-3 specifications. I think it could be interesting to get closer to standards on such aspect.

Pros:
- Keeps 100% compatibility with existing code
- Standardized logging, increased compatibility with other libraries
- Adds the notion of log level (but without any filtering possible at that time)

Cons:
- Adds [psr/log](https://packagist.org/packages/psr/log) as a Composer dependency

Below is a test case for the new implementation:

``` php
require 'vendor/autoload.php';
require 'log.php';

$logger = new Log('test.log');

$logger->debug('debug');
$logger->warning('warning {foo}', ['foo' => 'bar']);
$logger->emergency(
    'emergency {exception}',
    ['exception' => new Exception("it's a test"), 'unused_context_key' => 'nothing']
);

$logger->log('invalid_level', 'my message');
```

Result in `test.log`:

```
Wed, 06 Jul 2016 13:23:16 +0200 DEBUG debug
Wed, 06 Jul 2016 13:23:16 +0200 WARNING warning bar
Wed, 06 Jul 2016 13:23:16 +0200 EMERGENCY emergency Exception: it's a test in /var/www/fatfree-core/test.php:27
Stack trace:
#0 {main}
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,7 @@
 	"license": "GPL-3.0",
 	"require": {
 		"php": ">=5.3.6",
-        "psr/log": "1.0.0"
+		"psr/log": "1.0.0"
 	},
 	"autoload": {
 		"classmap": ["."]
--- a/composer.json
+++ b/composer.json
@@ -4,7 +4,8 @@
 	"homepage": "http://fatfreeframework.com/",
 	"license": "GPL-3.0",
 	"require": {
-		"php": ">=5.3.6"
+		"php": ">=5.3.6",
+        "psr/log": "1.0.0"
 	},
 	"autoload": {
 		"classmap": ["."]
--- a/log.php
+++ b/log.php
@@ -21,7 +21,9 @@
 */
 
 //! Custom logger
-class Log {
+class Log extends \Psr\Log\AbstractLogger {
+
+	protected static $levels;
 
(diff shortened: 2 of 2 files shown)


KOTRET:
the con-argument is a bummer, as F3 does not depend on any other libs. The dependency is not only added with composer, but also without using composer. I wonder if there is any chance to resolve this issue?

nanawel:
This is the cost of standards, I knew it was going to be the main issue here.
The only alternative I see is to provide the `LoggerInterface` interface (and only this) with F3, with a special check in case the interface has been declared before. Although I cannot be sure that it will work in all possible situations.

eazuka:
This whole standards thing tend to make things unnecessarily complex and build-up unnecessarily large base. For me, the PSR-3 specifications feature sets (extra methods) is nice but just like @KOTRET said, the  psr/log files dependency and Composer dependency to install is a no no.

nanawel:
@eazuka Point made. The developer is still free to choose whether or not he/she wants to use F3 `\Log` class or a different library.
Personally I use Monolog because it much more flexible for a bigger application with different contexts.

Issue #143, opened 2016-06-27
Combined operator doesn't update SQL mapper


xfra35:
If we're having an integer field:
* `$mapper->foo = $mapper->foo + 1` triggers an `UPDATE` statement
* `$mapper->foo++` triggers an `UPDATE` statement
* `$mapper->foo += 1` does not

The same issue happens with combined concatenation operator: `$mapper->bar.=$string`.

After digging up a bit in the code, I found out that [this line](https://github.com/bcosca/fatfree-core/blob/0083673bea46fd2c6fd03fdaf32666824b0f1c00/db/sql/mapper.php#L93) returns FALSE. In other words, when `set()` is called, `$this->fields[$key]['value']` equals `$val`, so the field doesn't get marked as "changed".

I'm scratching my head to understand what's going on here... Looks like there's some reference assignment messing up...

**UPDATE**
That definitely looks like something with references because that bug showed up while upgrading a project from 3.2 to 3.5. In 3.2, `Magic->get()` was not returning a reference.

ikkez:
I had a similar issue in cortex once ago. It's true that the reference is already updated when we process the magic setter. I worked around it which a hook in the `ArrayAccess` setter. 

Option A could be to add this method:

```php
public function offsetSet($key, $val) {
	$this->fields[$key]['changed']=TRUE;
	parent::offsetSet($key,$val);
}
```
and add a false case for the condition [here](https://github.com/bcosca/fatfree-core/blob/0083673bea46fd2c6fd03fdaf32666824b0f1c00/db/sql/mapper.php#L93) to invalidate the changed flag, so it becomes:

```php
if ($this->fields[$key]['value']!==$val ||
	$this->fields[$key]['default']!==$val && is_null($val))
	$this->fields[$key]['changed']=TRUE;
else
	$this->fields[$key]['changed']=FALSE;
```

Another way I had in mind is to extend the condition and add another check with the (original?) mapper within the `$query` property:

```php
if ($this->fields[$key]['value']!==$val ||
	$this->query[$this->ptr]->fields[$key]['value']!==$val ||
	$this->fields[$key]['default']!==$val && is_null($val))
	$this->fields[$key]['changed']=TRUE;
```

This might look like an easier fix now, but actually I am wondering why the mapper properties in the $query array isn't already referenced and updated to match the active loaded record on `$this`.
[...]

PR #150, opened 2016-08-26
Fixing issue #143 using an 'old_value' for each field.

Mindiell:
My solution for issue #143 : maybe not ideal because it fixes only the += operator. But it's quite simple ;o)


ikkez:
fixed

PR #151, opened 2016-08-26
Issue 143 bis

Mindiell:
This second PR is based on PR 150 : https://github.com/bcosca/fatfree-core/pull/150

It fixes the '++' operator using the old_value and creating a specific "save" method in the sql.php (so has to be added in the other specialized cursors). It verifies if value has been changed before saving.
This is because the '++' operator do not trigger any of the magic methods.


ikkez:
fixed in a similar way, thx for your investigation

Issue #145, opened 2016-07-21
Question about mapper "changed" flag


xfra35:
I just realized that when updating a mapper, the "changed" flag are not reset, which can lead to unexpected behaviour when running non-idempotent "before/afterupdate" hooks.

See this example:
```php
$mapper->email='foo@test.com';
$mapper->save();// afterupdate: if ($mapper->changed('email')) send email
$mapper->foo='bar';
$mapper->save();// email is sent twice
```

So should we consider this as a bug or as a misusage?

If we consider this is a bug, we should reset the "changed" flags inside `update()`.

KOTRET:
`[x]` bug
`[ ]` feature

saving an entity imho should mark the newly updated / inserted as unchanged

bcosca:
I agree this is a bug

ikkez:
fixed in https://github.com/bcosca/fatfree-core/commit/18ce506100d63ad852da78d877696d95f2e64995

PR #152, opened 2016-08-26
Issue #145: After update, 'changed' flag is set to false.

Mindiell:
Changed flag is set to FALSE for each field with changed flag set to TRUE.


ikkez:
fixed

PR #158, opened 2016-10-11
extend method

ikkez:
Here is a little proposal for a new extend method.
It's similar to the `merge` method, but only copies not existing keys. This way you can easily merge default values into existing hive array variables.

```php
// actual var
$f3->set('vars',['foo'=>'xyz', 2, 1]);
// apply defaults
$f3->extend('vars',[1,2,'cherry','bar'=>'baz','foo'=>'abc']);

Array (
    [foo] => xyz
    [0] => 2
    [1] => 1
    [2] => cherry
    [bar] => baz
)
```

using `merge` would lead to:

```php
$f3->set('vars',['foo'=>'xyz', 2, 1]);
$f3->merge('vars',[1,2,'cherry','bar'=>'baz','foo'=>'abc'], TRUE);

Array (
    [foo] => abc
    [0] => 2
    [1] => 1
    [2] => 1
    [3] => 2
    [4] => cherry
    [bar] => baz
)
```

### DIFF
--- a/base.php
+++ b/base.php
@@ -597,6 +597,20 @@ final class Base extends Prefab implements ArrayAccess {
 	}
 
 	/**
+	 *	Extend hive array variable with default values from $src
+	 *	@return array
+	 *	@param $key string
+	 *	@param $src string|array
+	 **/
+	function extend($key,$src) {
+		$ref=&$this->ref($key);
+		if (!$ref)
+			$ref=[];
+		$ref+=is_string($src)?$this->hive[$src]:$src;
+		return $ref;
+	}
+
+	/**
 	*	Convert backslashes to slashes
 	*	@return string
 	*	@param $str string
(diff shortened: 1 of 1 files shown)


xfra35:
That makes sense. Maybe keeping the same signature as in `merge()`? (`$keep=FALSE`)

ikkez:
Done. Looks very similar to `merge` now. Keep it as extra method or refactor it to be an extra parameter in `merge` instead? i.e. `merge($key,$src,$extend=FALSE,$keep=FALSE)` ??

xfra35:
I suggest to keep them separated, as they don't solve the same issues.

**Issue 1: apply defaults to an associative array**

```php
$f3->set('data',$input); // ['name'=>'foo']
$f3->extend('data',$defaults); // ['name'=>'foo', 'color'=>'blue']
```

**Issue 2: append data series**

```php
$f3->merge('data',$series1); // [1,2,3]
$f3->merge('data',$series2); // [1,2,3,3,5,2]
$f3->merge('data',$series3); // [1,2,3,3,5,2,9,5,3]
```

Issue #92, opened 2015-10-21
drop mset for LOCALES


ikkez:
Regarding this issue from the board https://groups.google.com/forum/#!topic/f3-framework/eDuCGUR1dBA we should think about not using `mset` here https://github.com/bcosca/fatfree-core/blob/master/base.php#L356 because `set` includes an existence check in the cache, that is triggered for each and every dictionary entry, which makes it slow for bigger lexicon files and is not really needed at all?!
I suggest to change the code to add lang keys in a more dumb way. maybe something like this:
```php
case 'LOCALES':
	if (isset($lex) || $lex=$this->lexicon($val))
		foreach ($lex as $key=>$val) {
			$ref=&$this->ref($this->hive['PREFIX'].$key);
			$ref=$val;
			unset($ref);
		}	
```

xfra35:
That `Cache->exists()` seems to be costing at times, even with `CACHE` disabled. I came to this conclusion [here](https://github.com/bcosca/fatfree/issues/845#issuecomment-113659667), although I didn't understand how it could slow things down with a [simple](https://github.com/bcosca/fatfree-core/blob/master/base.php#L2204) `if (!$this->dsn) return FALSE`.

By the way, what's the purpose of that check? Can't we just remove it?
```php
if ($ttl)
  Cache::instance()->set($hash,$val,$ttl);
```
looks fine to me..

Or maybe that check is there to give the possibility to invalidate a cache entry? Like:
```php
$f3->set('foo,'bar',3600);
$f3->set('foo','baz',0);//invalidate cache entry
```
In which case, we could maybe find an alternative to this? Like a 4th `$force` parameter, or a special `-1` value.

bcosca:
`if (!$this->dsn) return FALSE;` is simply for checking if `CACHE` is disabled.

ikkez:
I think this [line](https://github.com/bcosca/fatfree-core/blob/master/base.php#L370)
`if ($cache->exists($hash=$this->hash($key).'.var') || $ttl)`
updates the cache entry, using its original expire time. From the [API docs](http://fatfreeframework.com/base#Cachingproperties):
> If the $ttl parameter is > 0, and the framework cache engine is enabled, the specified variable will be cached for $ttl seconds. Already cached vars will be updated by reusing the old expiration time.

[...]

PR #159, opened 2016-10-12
drop mset for locales to optimize lang key loading #92

ikkez:
Here is a working approach that speeds up adding locales **x4**

Testes with 3 languages and ~150 language keys. Before:
```time: 39,77ms,  memory: 1048,6```

after this change:
```time: 10,3ms,  memory: 1048,6```

This is faster because we simply skip all things that `set` does. We can do this, because dictionary files are actually no config files and we don't expect special config variables in there (which `set` is checking for every time) and since the lexicon file is cached via `$ttl` in the whole, we don't need the caching checks in `set` at all. That'll give a boost.

### DIFF
--- a/base.php
+++ b/base.php
@@ -366,7 +366,11 @@ final class Base extends Prefab implements ArrayAccess {
 			$lex=$this->lexicon($this->hive['LOCALES'],$ttl);
 		case 'LOCALES':
 			if (isset($lex) || $lex=$this->lexicon($val,$ttl))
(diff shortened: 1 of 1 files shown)


xfra35:
Smart :+1: 

And that fixes the double caching issue.

ikkez:
I see this was fixed in https://github.com/bcosca/fatfree-core/commit/4ca93748850319b875a7185fda510fe62b95a0f4

PR #163, opened 2016-10-13
Adding CORS Force Ability

chez14:
Force enabling CORS for development purposes.
The current behaviour of the CORS support are checking the `HEADERS.Origin`, if it exist, then the CORS support are enabled. 

By adding the `force` var, it will always activate the CORS support, no matter if the  `HEADERS.Origin` are exist or not.

Using:

``` php
// This will enable the CORS Support forcefully.
\F3::set('CORS.force', true);
```

### DIFF
--- a/base.php
+++ b/base.php
@@ -1451,7 +1451,7 @@ final class Base extends Prefab implements ArrayAccess {
 		$this->hive['ROUTES']=array_combine($keys,$vals);
 		// Convert to BASE-relative URL
 		$req=$this->rel(urldecode($this->hive['PATH']));
-		if ($cors=(isset($this->hive['HEADERS']['Origin']) &&
+		if ($cors=((isset($this->hive['HEADERS']['Origin']) || $this->hive['CORS']['force']) &&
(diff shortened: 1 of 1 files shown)


ikkez:
> The current behaviour of the CORS support are checking the HEADERS.Origin, if it exist, then the CORS support are enabled. 

Well actually this is how CORS works.

> https://www.w3.org/TR/cors/
> Server-side applications are enabled to discover that an HTTP request was deemed a cross-origin request by the user agent, through the Origin header.

>https://www.w3.org/TR/cors/#resource-requests
> Resources must use the following set of steps ... :
> 1. If the Origin header is not present terminate this set of steps. The request is outside the scope of this specification.

When you enable the CORS headers at your client application, you shouldn't have a need for this `force` feature. i.e. for jQuery, it is:

```js
$.ajaxSetup({
	crossDomain: true
});
```

chez14:
Ohh.. ok, sorry hhehehehe, i'll close this...

PR #166, opened 2016-10-17
NEW: Sql ttl arrays

xfra35:
This PR is a follow up of [this thread](https://groups.google.com/forum/#!topic/f3-framework/t0hYxPVuFeE).

It reverts https://github.com/bcosca/fatfree-core/commit/3122b5881ee792f4ab3ad71522444f1900e7820c and gives instead the possibility to tag cached queries, so that it becomes possible to selectively reset a subset of all the cached queries.

The change is achieved through ttl arrays in order to keep BC:

``` php
// controller1
$db->exec($sql,$args,[$ttl,'product']); // <-- tag that query as 'product' related
$mapper->find($filter1,$opts,[$ttl,'product']); // idem

// controller2
$mapper->find($filter2,$opts,[$ttl,'product']); // idem

// controller3
$mapper->find($filter3,$opts,[$ttl,'customer']); // different tagging here

// controller 4 (clear 'product' related cached queries)
$cache->reset('.product.sql');
```

### DIFF
--- a/db/sql.php
+++ b/db/sql.php
@@ -46,9 +46,7 @@ class SQL {
 		//! Number of rows affected by query
 		$rows=0,
 		//! SQL log
-		$log,
-		//! last used hash keys
-		$hash=[];
+		$log;
 
 	/**
 	*	Begin SQL transaction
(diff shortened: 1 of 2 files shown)


ikkez:
If we extend this solution also to the Cache class, it would probably solve https://github.com/bcosca/fatfree/pull/674/files as well.

xfra35:
Fixing that other issue requires to move the whole hashing/suffixing stuff into one central place (the `Cache` class).

We could tweak the `$key` inside the `Cache` class:

```php
$cache->set('foo','bar',90); // hash 'foo'
$cache->set(['foo','.suffix'],'bar',90); // hash 'foo' and append '.suffix'
echo $cache->get(['foo','.suffix']); // bar
```

then the `$hash`  computed inside `DB\SQL->exec()` would become a plain `$key`:
```php
$key=[$this->dsn.$cmd.$fw->stringify($arg),
  ($tag?'.'.$tag:'').'.sql'];
```

and finally, the `Cache` hashing function would be customizable:
```php
$cache->hash(function($key){
  return Web::instance()->slug($key);
});
```

ikkez:
well, that sounds pretty neat, isn't it?

eazuka:
Nice +1

Issue #106, opened 2015-12-18
PDO string conversion of floating values not reliable


xfra35:
Since PDO doesn't handle floating points, the framework has to [cast](https://github.com/bcosca/fatfree-core/blob/c9681ed5b2b5ec34abca1afd9be275d469aa24ec/db/sql.php#L116) float values to string.

This work well with locales having `.` as a decimal separator, but not for those having `,`:
```php
$f3->LANGUAGE='fr';
$price=12.34;
$mapper->price=$price;
var_dump($mapper->price);// '12,34' <= there shouldn't be a comma here
$mapper->save();// <= value stored in database = 12
```
The only way to have this working is to perform a custom string cast before each call to `set` or to overload the `set()` method:
```php
$mapper->price=str_replace(',','.',$price);
var_dump($mapper->price);// '12.34' <= OK
```
This is not convenient and can lead to mistakes.

I think that the framework should handle it, but I'm not sure how. Maybe by adding a `float` flag in the `$db->schema` result?

Any idea?

geniuswebtools:
I think the solution for this would be in the beforeSave() trigger.  The reason I say that is because saving the result of 12 is most likely the result of the database engine and the data type associated with that column.  

I'll assume you're not using SQLite because I'd wager that the SQLite would save the 12,34 value as passed because it doesn't truncated values like the MySQL decimal type does.

ikkez:
So is this a number format/cast issue related to the system locale? I think the mapper only handles floats as string because the topic can be quite complex and full of traps:
* float/integer limitation on 32bit systems
* there is not native `PDO::PARAM_FLOAT`
* possibly rounding issues
* casting locale numbers with thousand separators and decimals can be tricky:
  * DE: `1.234,99`
  * EN: `1,234.99`

But if you set a float value to a mapper and save it to a float or decimal db table column, it should not be trimmed. Is that due to your `fr` language setting?

xfra35:
@geniuswebtools is right, the value gets truncated because MySQL expects a dot as a separator, not a comma. I haven't tried yet with other SQL engines.

> Is that due to your fr language setting?

Yes, and I expect the same behaviour with `de`. Haven't you face the problem already?

ikkez:
Tested. You are right. When the float value is casted to a string, it depends on the current system locale how it's going to be formatted. I suggest to use a custom reliable format here:

```php
$mapper->price = number_format(2234.99, 2, '.', '');
```

[...]

PR #167, opened 2016-10-26
add float support for SQL DB #106

ikkez:
Here is a fix for handling float assignments in the SQL class. It appeared that the actual problem doesn't was the assignment as string, but the type conversion to string, due to the missing native cast feature in the `value()` method to a float. Bundled with a little string sanitation, this should properly cast to native float values now and be stored correctly in the DB. (Yes `->bindvalue` will receive a float value and set is as `PDO::PARAM_STR`. Surprisingly this was the only proper solution).

### DIFF
--- a/db/sql.php
+++ b/db/sql.php
@@ -30,6 +30,9 @@ class SQL {
 		E_PKey='Table %s does not have a primary key';
 	//@}
 
+	const
+		PARAM_FLOAT='float';
+
 	protected
 		//! UUID
 		$uuid,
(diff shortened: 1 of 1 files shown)


xfra35:
Thanks @ikkez. That fix looks perfect :+1:

Just one small question: what's the `preg_replace('/([.,])(?!\d+$)/','',$val)` for? Is this for handling the case where there would be a trailing separator? In that case, this looks superfluous as both `(float)'12.'` and `(float)'12,'` trim that character.

ikkez:
It removes all `.` and `,` but not the last one. It's meant to ensure that string `12.345.678,99` becomes float `12345678.99`, regardless of the used locale.

xfra35:
OK I got it. Actually this patch fixes #106 but also adds a new feature, which is the possibility to assign a formatted string to a float/decimal field:

```php
$mapper->price='1,200.34'; // works!
```

PR #169, opened 2016-11-12
Correct handling of TTL for Redis

crypto-scythe:
If TTL is 0 the 'ex' flag is not setting the key as it would instantly expire. To create a key without TTL we need to omit the 'ex' flag.


ikkez:
What's the default for the 3rd parameter? does this work as well:
```php
return $this->ref->set($ndx,$data,$ttl ? [ex'=>$ttl] : []);
```
?

crypto-scythe:
Yeah you're right that works as well, I just updated the code :)

PR #180, opened 2016-12-21
Fix locales precedence order and charset hyphen

xfra35:
I submit these 2 fixes as a PR, because I'm surprised that it didn't come earlier (so I may be missing something).

Fix 1:
Locales precedence order: I believe that `fr_FR.UTF-8` should be processed before `fr_FR`.

Fix 2:
The replacement of hyphens by underscores should not affect the charset:
```
fr-FR.UTF-8 => fr_FR.UTF-8
```

### DIFF
--- a/base.php
+++ b/base.php
@@ -1009,10 +1009,11 @@ final class Base extends Prefab implements ArrayAccess {
 					$country=@constant('ISO::CC_'.strtolower($parts[1])))
 					$locale.='-'.$country;
 			}
-			$locales[]=$locale;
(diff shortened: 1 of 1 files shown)


ikkez:
ah yes. that's probably the reason why i sometimes had issues with the language auto-detection :D

xfra35:
Funny how I didn't notice it either, although I use it all the time. Probably because on most servers, the locale without charset links to the UTF-8 locale. But today, the setup was different, and that's how I noticed the bug.

PR #182, opened 2017-01-10
Return only the client's IP when using the X-Forwarded-For header

Rayne:
This PR fixes https://github.com/bcosca/fatfree/issues/985 by extracting the client's IP (first comma separated segment) from the `X-Forwarded-For` header.

### DIFF
--- a/base.php
+++ b/base.php
@@ -1163,7 +1163,7 @@ final class Base extends Prefab implements ArrayAccess {
 		return isset($headers['Client-IP'])?
 			$headers['Client-IP']:
 			(isset($headers['X-Forwarded-For'])?
-				$headers['X-Forwarded-For']:
+				explode(',',$headers['X-Forwarded-For'])[0]:
 				(isset($_SERVER['REMOTE_ADDR'])?
 					$_SERVER['REMOTE_ADDR']:''));
 	}


base.php
```diff
@@ -1163,7 +1163,7 @@ function ip() {
 		return isset($headers['Client-IP'])?
 			$headers['Client-IP']:
 			(isset($headers['X-Forwarded-For'])?
-				$headers['X-Forwarded-For']:
+				explode(',',$headers['X-Forwarded-For'])[0]:
```
xfra35: Maybe a `trim`? In bcosca/fatfree#985, you've stated:

>  the value is a comma+space separated list of IP addresses

kavin-90:
This can be done by SERVER side too. Websever provide REAL IP plugin inbuilt so it can get real ip even if you may try proxy .

base.php
Rayne: I also thought about it, but the schema is `, ` and luckily not ` ,` for separating IP addresses. I hope all servers respect this rule.

Rayne:
@kavin-90 Yes, it's also possible on the server side. Please read this post why I created the PR: https://github.com/bcosca/fatfree/issues/985#issuecomment-270003333

base.php
xfra35: Ah true. I forgot that we need only the first value.

PR #191, opened 2017-02-08
DB\SQL::PARAM_FLOAT: remove cast to float (#106 and bcosca/fatfree#984)

xfra35:
Cf. https://github.com/bcosca/fatfree/issues/984#issuecomment-278477598

### DIFF
--- a/db/sql.php
+++ b/db/sql.php
@@ -49,9 +49,7 @@ class SQL {
 		//! Number of rows affected by query
 		$rows=0,
 		//! SQL log
-		$log,
-		//! float precision break-point
-		$precision;
+		$log;
 
 	/**
 	*	Begin SQL transaction
@@ -122,12 +120,7 @@ class SQL {
 	function value($type,$val) {
 		switch ($type) {
 			case self::PARAM_FLOAT:
-				if (is_string($val)) {
-					$val=str_replace(',','.',preg_replace('/([.,])(?!\d+$)/','',$val));
-					if (strlen($val)-(int)(bool)strpos($val,'.') <= $this->precision)
-						$val=(float) $val;
-				}
-				return $val;
+				return rtrim(sprintf('%F',$val),'0');
 			case \PDO::PARAM_NULL:
 				return (unset)$val;
 			case \PDO::PARAM_INT:
@@ -375,7 +368,7 @@ class SQL {
 											'/blob|bytea|image|binary/i',
 											$row[$val[2]])?\PDO::PARAM_LOB:
 											(preg_match(
-												'/float|double/i',
+												'/float|decimal|real|numeric|double/i',
 												$row[$val[2]])?self::PARAM_FLOAT:
 												\PDO::PARAM_STR))),
 							'default'=>is_string($row[$val[3]])?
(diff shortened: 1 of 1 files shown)


db/sql.php
```diff
@@ -122,12 +120,7 @@ function type($val) {
 	function value($type,$val) {
 		switch ($type) {
 			case self::PARAM_FLOAT:
-				if (is_string($val)) {
-					$val=str_replace(',','.',preg_replace('/([.,])(?!\d+$)/','',$val));
-					if (strlen($val)-(int)(bool)strpos($val,'.') <= $this->precision)
-						$val=(float) $val;
-				}
-				return $val;
+				return rtrim(sprintf('%F',$val),'0');
```
Rayne: That line returns `n.` instead of `n` when `$val = n` and `n` is an integer. Is this the expected behaviour?

`$val` | Result
-|-
`1` | `1.`
`1.0` | `1.`
`1.10` | `1.1`

xfra35: Actually it is a valid format and `1.` makes it clear that it's a float (and not an integer).  But I guess we don't need it. So I've just switched `sprintf` with a simple `str_replace`.

Also I've restricted the processing to `float` only, so that values coming from the db remain untouched.

xfra35:
According to this PR, the expected behaviour would be, considering we have a field `num` of type `FLOAT` or `DECIMAL` in database:

```php
var_dump($mapper->num);// string(6) "12.340"
$mapper->num=13.50;
var_dump($mapper->num);// string(5) "13.50"
```
..no matter which locale is set.

If you're working with high precision numbers, you'd better pass a string directly:
```php
$mapper->num='13.12345678901234567890';
var_dump($mapper->num);// string(23) "13.12345678901234567890"
```
..otherwise some precision may be lost:
```php
$mapper->num=13.12345678901234567890;
var_dump($mapper->num);// string(15) "13.123456789012"
echo ini_get('precision');// 14
```

xfra35:
Are you sure we should take care of this? The value coming from an `<input>`  should be validated first and converted to a float or a "float string" depending on the precision.

We don't expect the following to work:

```php
$a='1 234,56';
echo 2*$a;
```

so why should this work?
```php
$mapper->a='1 234,56';
echo 2*$mapper->a;
```

Anyway, if we do it, let's include space separators and trim the input first to avoid side effects:
```php
$val=str_replace(',','.',preg_replace('/([., ])(?!\d+$)/','',trim($val)));
```

ikkez:
Well I'm just saying, that it could become an issue. Maybe when using `<input type="number" step="0.01">` and a simple `$mapper->copyfrom('POST');`. Depending on the browser locale or frontend javascript applied to input fields, the incoming values could include such glitches.
And this behaviour was [already included in the first fix](https://github.com/bcosca/fatfree-core/pull/167#issuecomment-256596073)

xfra35:
I remember that it was included in the first fix, but now I realize that parsing all possible number formats is a totally different job, that requires extra processing.

The regex cannot distinguish between English `1,234` (meaning `1234`) and Canadian `1,234` (meaning `1.234`). Or between English `1.234` and Spanish `1.234`. Cf. https://docs.oracle.com/cd/E19455-01/806-0169/overview-9/index.html

The only reason why we have to replace commas by points is because of PHP's specific way of converting floating points to strings: it includes the locale decimal point, but discards the thousand separator. It's like a half job done:

```php
setlocale(LC_ALL,'en_US.UTF-8');
echo 1234.56;// 1234.56 although the locale format is 1,234.56
setlocale(LC_ALL,'it_IT.UTF-8');
echo 1234.56;// 1234,56 although the locale format is 1.234,56
```

ikkez:
Alright, then there is probably no easy way to guess the right value from a formatted string and is out of the mapper's responsibility.

PR #193, opened 2017-02-21
NEW: quick erase flag on Mongo/SQL mappers

xfra35:
Here's a proposal to add a parameter to the mappers `erase()` method, so that we can choose one of the two following behaviors:

**Quick erase:** (default)
Delete all records at once and skip before/aftererase hooks.

```php
$mapper->erase($filter);
```

**Slow erase:**
Delete and trigger before/aftererase hooks for each record one by one.

```php
$mapper->erase($filter,FALSE);
```

NB: Jig mapper does only offer the 2nd behaviour. We should maybe implement the first one (quick erase), for consistency purposes.

### DIFF
--- a/db/mongo/mapper.php
+++ b/db/mongo/mapper.php
@@ -292,10 +292,17 @@ class Mapper extends \DB\Cursor {
 	/**
 	*	Delete current record
 	*	@return bool
+	*	@param $quick bool
 	*	@param $filter array
 	**/
-	function erase($filter=NULL) {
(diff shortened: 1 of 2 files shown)


ikkez:
Implemented on Jig mapper, though the default is `$quick=FALSE` there in order to keep the backward compatibility. Can could unify that in a next major version.

xfra35:
> We could unify that in a next major version

Agree. I think `$quick=FALSE` could be the common default behaviour, as one would expect hooks to be triggered on every operation.

PR #194, opened 2017-02-21
NEW: Recaptcha plugin

xfra35:
Usage:

1) Request API keys for your domain [here](https://www.google.com/recaptcha/admin?hl=en). You should get two keys: a site key and a secret key. NB: you can request keys for `localhost` or IP addresses as well.
2) Add the reCAPTCHA widget to your page:
    * `<div class="g-recaptcha" data-sitekey="YOUR_SITEKEY"></div>`
    * `<script src='https://www.google.com/recaptcha/api.js'></script>`
3) Validate the response on the server side with the `Recaptcha` plugin:
```php
if (!Web\Google\Recaptcha::verify('YOUR_SECRETKEY')) {
  // reCAPTCHA validation failed
}
```

I still prefer the native captcha feature coming with the framework, as it doesn't require any external javascript nor request, but since I needed reCAPTCHA on a project, I thought I'd share it.

NB: the plugin is a fat-free refactoring of Google's [own plugin](https://github.com/google/recaptcha).

### DIFF
--- a/web/google/recaptcha.php
+++ b/web/google/recaptcha.php
@@ -1,0 +1,57 @@
+<?php
+
+/*
+
+	Copyright (c) 2009-2017 F3::Factory/Bong Cosca, All rights reserved.
+
+	This file is part of the Fat-Free Framework (http://fatfreeframework.com).
+
+	This is free software: you can redistribute it and/or modify it under the
+	terms of the GNU General Public License as published by the Free Software
+	Foundation, either version 3 of the License, or later.
+
+	Fat-Free Framework is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+	General Public License for more details.
+
+	You should have received a copy of the GNU General Public License along
+	with Fat-Free Framework.  If not, see <http://www.gnu.org/licenses/>.
+
+*/
+
+namespace Web\Google;
+
+//! Google ReCAPTCHA v2 plug-in
+class Recaptcha {
+
+	const
+		//! API URL
+		URL_Recaptcha='https://www.google.com/recaptcha/api/siteverify';
+
+	/**
+	 *	Verify reCAPTCHA response
+	 *	@param string $secret
+	 *	@param string $response
+	 *	@return bool
+	 **/
+	static function verify($secret,$response=NULL) {
+		$fw=\Base::instance();
+		if (!isset($response))
+			$response=$fw->get('POST.g-recaptcha-response');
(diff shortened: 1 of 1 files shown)


web/google/recaptcha.php
```diff
@@ -0,0 +1,57 @@
...
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+	General Public License for more details.
+
+	You should have received a copy of the GNU General Public License along
+	with Fat-Free Framework.  If not, see <http://www.gnu.org/licenses/>.
+
+*/
+
+namespace Web\Google;
```
KOTRET: maybe choose another namespace like `F3\Plugins\Google`?

kavin-90:
I use official **google/recaptcha**  package 😄

bcosca:
This may require some tweaking. Consider this possibility:

1. Form is displayed via GET method.
2. Form is submitted via POST method.
3. Some entries are invalid so the same form is redisplayed with alerts/messages (still using POST).
4. Try refreshing/reloading the page. Oops, we have a reCaptcha error.

kavin-90:
yeah there is reload recaptcha function in js which reload recaptcha.

xfra35:
The command is `grecaptcha.reset()`.

ikkez:
The recaptcha service doesn't neccessarily need a SESSION interaction. Instead, the google recapcha API will tell us based on the `POST.g-recaptcha-response` token, that was generated in the frontend by the recaptcha javascript, if that user validation was successful. If you had form errors and need to display the form again, you have to solve the captcha again.

https://yuml.me/4e782e71

Sweet little plugin. Maybe too more methods would be handy:

1. something like `js()` that will return the public path to the javascript file:`https://www.google.com/recaptcha/api.js`
2. and maybe `widget($siteKey)` that will return the snippet for the form integration: `<div class="g-recaptcha" data-sitekey="'.$siteKey.'"></div>`, so we don't need to look this up all again

kavin-90:
Example verification of recpatcha

```
 $peer_key = version_compare(PHP_VERSION, '5.6.0', '<') ? 'CN_name' : 'peer_name';
        $options = array(
            'http' => array(
                'header' => "Content-type: application/x-www-form-urlencoded\r\n",
                'method' => 'POST',
                'content' => $params->toQueryString(),
                // Force the peer to validate (not needed in 5.6.0+, but still works)
                'verify_peer' => true,
                // Force the peer validation to use www.google.com
                $peer_key => 'www.google.com',
            ),
        );
        $context = stream_context_create($options);
        return file_get_contents(self::SITE_VERIFY_URL, false, $context);
```

PR #197, opened 2017-02-28
Fix "Undefined index: CLI" error

nanawel:
Commit 894c3ea9beab2ea77e08e2f116d94ca58a6f73d2 seems to have introduced a bug that makes the framework crash when attempting to render a template because `$hive['CLI']` does not exist.
I'm proposing this patch to fix this issue.

### DIFF
--- a/base.php
+++ b/base.php
@@ -2685,7 +2685,7 @@ class View extends Prefab {
 			$hive=$fw->hive();
 		}
 		if ($this->level<1 || $implicit) {
-			if (!$hive['CLI'] && !headers_sent() &&
+			if ((!isset($hive['CLI']) || !$hive['CLI']) && !headers_sent() &&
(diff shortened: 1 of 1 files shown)


xfra35:
`$fw->get('CLI')` is more appropriate as it refers to the framework's hive

nanawel:
@xfra35 That could be better but not in that case, as `$hive` can also be an argument of the method.

xfra35:
Precisely, `CLI` in that case refers to the global hive. It's just here to detect CLI mode. You could replace it with `PHP_SAPI!='cli'` but that wouldn't cover the `mock()` method. In any case, it hasn't anything to do with the template hive.

nanawel:
Well, the original modification was not mine so I prefered to keep the idea. But okay it's fine for me, I made the modification you proposed. Thanks.

PR #198, opened 2017-02-28
View: Don't delete sandboxed variables

Rayne:
References:

* https://github.com/bcosca/fatfree-core/pull/46
* 88246a45e684991f4074e0a98532c4270030e362


Rayne:
The following pseudo code allows us to clean the sandbox (local template variables) from all unexpected variables:

```php
class View {
    private $tmpHive;

    public function sandbox(array $hive = null, $mime = null) {
        // …
        $this->tmphive = $hive;
        unset($fw,$hive,$implicit,$mime) // Unset all local variables.
        extract($this->tmpHive);
        // …
    }
}
```

I would like to create a PR for this clean solution. What's your idea or opinion?

ikkez:
Sounds reasonable to me. You could also add another commit to this (your) branch here and add it directly.

PR #202, opened 2017-03-25
Fix bcosca/fatfree#1023: Accept quoted strings in language .ini files

kumy:


ikkez:
Unfortunately this doesn't fit well here and breaks existing things. See comment on #1023

PR #209, opened 2017-04-12
SQL Server Information Schema Mapper Issue

EatZeBaby:
Proposal for a fix an issue when querying the information schema on SQL Server.

The issue was PDOStatement: [Microsoft][ODBC Driver 11 for SQL Server][SQL Server]Invalid object name 'information_schema.columns'.

### DIFF
--- a/db/sql.php
+++ b/db/sql.php
@@ -317,7 +317,36 @@ class SQL {
 			'mysql'=>[
 				'SHOW columns FROM `'.$this->dbname.'`.`'.$table.'`',
 				'Field','Type','Default','Null','YES','Key','PRI'],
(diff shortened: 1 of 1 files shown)


ikkez:
After asking google, it seems like there is indeed a case-sensitive collation existing for sql server. So rewriting the query in all uppercase for the information schema looks like a valid fix to me. I think it can be applied for `mssql|sqlsrv|sybase|dblib|pgsql|odbc` as well and shouldn't be harmful (at least a quick test on postgre worked just fine, and mssql, sqlsrv, odbc are basically just different drivers for sql server AFAIK).

bcosca:
I would be careful with this. SQL keywords are case-insensitive but name are not.

PR #211, opened 2017-04-21
Fix for issue: Template incorrectly setting header

sevn:
Setting mimetype to null prevents any headers being set.

### DIFF
--- a/base.php
+++ b/base.php
@@ -2690,9 +2690,9 @@ class View extends Prefab {
 			$hive=$fw->hive();
 		}
 		if ($this->level<1 || $implicit) {
-			if (!$fw->CLI && !headers_sent() &&
+			if (!$fw->CLI && $mime!=NULL && !headers_sent() &&
 				!preg_grep ('/^Content-Type:/',headers_list()))
-				header('Content-Type: '.($mime?:'text/html').'; '.
+				header('Content-Type: '.$mime.'; '.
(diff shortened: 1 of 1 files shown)


base.php
```diff
@@ -2690,9 +2690,9 @@ protected function sandbox(array $hive=NULL,$mime=NULL) {
 			$hive=$fw->hive();
 		}
 		if ($this->level<1 || $implicit) {
-			if (!$fw->CLI && !headers_sent() &&
+			if (!$fw->CLI && $mime!=NULL && !headers_sent() &&
```
KOTRET: `&& $mime &&` should be enough

sevn: Rest of the code follows in the minimalist style anyway so your comment is valid, I just liked to show it explicit because it has a strong functional effect.

ikkez:
It's about _not_ setting the header at all, when you pass `NULL` as `$mime` argument, so that the header is correct when you, i.e. render a bunch of email templates (text/plain) first (and set $mime to NULL), but then render something different for the actual response, i.e. a CSS file or text/html.

sevn:
I noticed that the mime type was set to 'text/html' on a conditional
statement when mime was null. So if one actually wanted a null mime type it
was not possible. This way one can set the mime type to null to prevent the
header being set altogether.

Plus it kind of makes more sense to have default values in the method
signature.

Issue #215, opened 2017-05-13
SMTP line length


ikkez:
The SMTP RFC https://tools.ietf.org/html/rfc5321#section-4.5.3.1 says that the text line should have a maximum length of 1000.

We should probably add a chunk split in the smtp plugin to handle that.

bcosca:
`chunk_split` may not be appropriate for cases like HTML email. I think `wordwrap` is more up to the job.

KOTRET:
Notice that wordwrap's `$cut`-param must be `TRUE`.
Are there sideeffects when encoding a file for transfer like unwanted splitting a chunk below 1000 length due to base64-chars that are treated as word boundary? What about `<pre>`-tags in html mails?

bcosca:
Chunked base64 encoding is a non-issue. It's quote-printable that may cause problems if text lines exceed the limit, but this can be controlled at the app end - not the SMTP plugin.

SylwesterZarebski:
It should be also noted, that 1000 characters (bytes) limit apply to lines send to SMTP server, not original ones, which could be transformed by quoted-printable encoding and grows up - this "expanding" could be easily seen with multi-byte UTF-8 words.

For quoted-printable encoding You could use soft line breaks - just = on end of line:
https://en.wikipedia.org/wiki/Quoted-printable - see Example.

PR #218, opened 2017-05-23
fix SMTP line width #215

ikkez:
Proposal fix for #215


SylwesterZarebski:
PHP wordwrap function is not multi-byte safe and will wrap inside UTF characters breaking them. You should use something like this:
https://stackoverflow.com/questions/3825226/multi-byte-safe-wordwrap-function-for-utf-8

PR #224, opened 2017-09-13
Attention Ikkez

sevn:
These includes changes discussed yesterday.

I thought about beforerender and what argument it should be passed. To me the logical thing to pass is the absolute path to the view to be rendered. This way you can do whatever you want to the html pre-rendering. Perhaps even prepre-rendering through a separate template engine?

### DIFF
--- a/base.php
+++ b/base.php
@@ -2690,9 +2690,9 @@ class View extends Prefab {
 			$hive=$fw->hive();
 		}
 		if ($this->level<1 || $implicit) {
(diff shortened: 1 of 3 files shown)


Rayne:
Please set a more specific title, add some details and rebase/squash the commits. There are many small changes but I think there should be a better way to name and group them.

sevn:
Absolutely. Apologies for my git newbness.

PR #225, opened 2017-09-16
fix sorting issues with sub-queries on SQL Server, bcosca/fatfree#1052

ikkez:
Some fixes to resolve https://github.com/bcosca/fatfree/pull/1052
@dexwerx I've checked your proposal and applied them with only slightly adjustments.
Tested on SQL Server2008 and 2012 here, as well as on the other engines just to be sure it breaks nothing else and it's looking good. I wasn't able to get sorting on an adhoc field + limit + offset working for sqlsrv 2008 though, but I start to believe that it's either not working in general for that specific version or it's my old version here.. got no problems on sqlsrv 2012 though.
Please check if that resolves your issue.

### DIFF
--- a/db/sql/mapper.php
+++ b/db/sql/mapper.php
@@ -348,7 +348,7 @@ class Mapper extends \DB\Cursor {
 		foreach ($this->adhoc as $key=>$field)
 			$adhoc.=','.$field['expr'].' AS '.$this->db->quotekey($key);
 		$fields='*'.$adhoc;
-		if (preg_match('/mssql|dblib|sqlsrv/',$this->engine) && (strncmp($this->db->version(),'11',2)<0))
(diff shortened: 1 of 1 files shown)


dexwerx:
Tested (sorting + limit + offset) on SQL Server 2008, and it works.
line 351 however breaks SQL Server 2012
> PDOStatement: [Microsoft][ODBC Driver 11 for SQL Server][SQL Server]The ORDER BY clause is invalid in views, inline functions, derived tables, subqueries, and common table expressions, unless TOP, OFFSET or FOR XML is also specified.

it does however work if i comment out the ` && (strncmp($this->db->version(),'11',2)<0))`

ikkez:
Tested now in SQL Server 2008, 2012 and 2014... you're hint is valid and I've fixed this in another commit. Looks good now.

PR #227, opened 2017-10-16
LDAP changes to for AD flexibility

docwyatt2001:
As mentioned (and similar to PR1068 on bcosca/fatfree), changes made to make the LDAP auth work against and Active Directory environment ,with a few changes to keep it backwardly compatible to original code.

### DIFF
--- a/auth.php
+++ b/auth.php
@@ -116,7 +116,6 @@ class Auth {
 	**/
 	protected function _ldap($id,$pw) {
 	  $port=intval($this->args['port']?:389)
-	  $this->args['filter']=$this->args['filter']?:"uid=".$id;
 	  $filter=$this->args['filter']=$this->args['filter']?:"uid=".$id;
 	  $this->args['attr']=$this->args['attr']?:["uid"];
 	  array_walk($this->args['attr'],
--- a/auth.php
+++ b/auth.php
@@ -115,20 +115,31 @@ class Auth {
 	*	@param $pw string
 	**/
 	protected function _ldap($id,$pw) {
-		$dc=@ldap_connect($this->args['dc']);
-		if ($dc &&
-			ldap_set_option($dc,LDAP_OPT_PROTOCOL_VERSION,3) &&
-			ldap_set_option($dc,LDAP_OPT_REFERRALS,0) &&
-			ldap_bind($dc,$this->args['rdn'],$this->args['pw']) &&
-			($result=ldap_search($dc,$this->args['base_dn'],
-				$this->args['uid'].'='.$id)) &&
-			ldap_count_entries($dc,$result) &&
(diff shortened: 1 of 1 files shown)


auth.php
```diff
@@ -115,20 +115,31 @@ protected function _sql($id,$pw,$realm) {
...
-			ldap_count_entries($dc,$result) &&
-			($info=ldap_get_entries($dc,$result)) &&
-			@ldap_bind($dc,$info[0]['dn'],$pw) &&
-			@ldap_close($dc)) {
-			return $info[0][$this->args['uid']][0]==$id;
-		}
-		user_error(self::E_LDAP,E_USER_ERROR);
+	  $port=intval($this->args['port']?:389)
+	  $this->args['filter']=$this->args['filter']?:"uid=".$id;
+	  $filter=$this->args['filter']=$this->args['filter']?:"uid=".$id;
```
KOTRET: 2nd assignment repeated from one line above

docwyatt2001: Damn - missed that. :(

ikkez:
Looks fine 👍 Can someone please verify/test this change once more?!.. got no LDAP instance here atm.

docwyatt2001:
It works on the project I am using it currently for, but to play it safe, I can fire up a new LDAP server after work, and test it against that, if that's acceptable.

Rayne:
The test cases are in the [fatfree/dev](https://github.com/bcosca/fatfree/tree/dev) branch. The problem is, that there are no LDAP tests (or I found none). I found only a few tests for [Auth](https://github.com/bcosca/fatfree/blob/dev/app/auth.php).

auth.php
```diff
@@ -115,20 +115,30 @@ protected function _sql($id,$pw,$realm) {
...
+	    ldap_set_option($dc,LDAP_OPT_REFERRALS,0) &&
+	    ldap_bind($dc,$this->args['rdn'],$this->args['pw']) &&
+	    ($result=ldap_search($dc,$this->args['base_dn'],
+	     $filter,$this->args['attr'])) &&
+	    ldap_count_entries($dc,$result) &&
+	    ($info=ldap_get_entries($dc,$result)) &&
+	    $info['count']==1 &&
+	    @ldap_bind($dc,$info[0]['dn'],$pw) &&
+	    @ldap_close($dc)) {
+	      return in_array($id,(array_map(function($value){
```
ikkez: is this array_map necessary? I think there's always just 1 result and as you are checking for `$info['count']==1` it maybe could be simplified here?!

jekabsmilbrets:
there is missing semicolon at row 118 at lib/auth.php

ikkez:
fixed. thanks

Issue #178, opened 2016-12-13
Add a 'JAR.lifetime' param to set expiration in seconds


sam2kb:
This code is from Base::set()
```
$jar=$this->unserialize($this->serialize($this->hive['JAR']));
$jar['expire']-=$time;
call_user_func_array('session_set_cookie_params',$jar);
```
Why would you subtract time, this makes JAR.expire value negative. Unless you add time() first, but why on Earth would one add time and then subtract it later.

ikkez:
From the quick reference [JAR](https://fatfreeframework.com/3.6/quick-reference#JAR)

> `expire` Unix timestamp, when the cookie should expire. Default: 0

So you had to set a unix timestamp to `JAR.expire`, which is usually greater than `time()`. In case of the session cookie parameter, this piece of code you've picked just converts the unix timestamp to a simple seconds-based lifetime value by substracting `$time`, which is necessary because the expire parameter in `setcookie` and `session_set_cookie_params` are different.

sam2kb:
That's right, one can't just set expiration time in seconds in the config file

[JAR]
expire = 604800
domain = whatever.tld

So my question is why do we require Unix timestamp instead of time in seconds like in session_set_cookie_params

sam2kb:
Ok I see now that `JAR.expire` is used elsewhere in the code as timestamp and we are not supposed to define this in text config.
Should we then have another `lifetime` param in seconds for use in config files?

PR #239, opened 2018-01-02
add JAR.lifetime option, #178

ikkez:
Here is a proposal for #178 to add a `lifetime` param for the session and cookie settings `JAR`.
We got `JAR.expire` so far, which is a timestamp that is used for 
`session_set_cookie_params` (needs ttl) and `setcookie` (needs timestamp)

As this statement was hard to express in a config file
 `$f3->set('JAR.expire', time()+604800);`
you can now use the `lifetime` parameter to set a lifetime in seconds instead of a timestamp:

```ini
[globals]
JAR.lifetime  = 3600
```

It's also possible to use a string time (strtotime is baked in), i.e.:

```ini
JAR.lifetime = 2 hours
JAR.lifetime = 3 days
JAR.lifetime = 1 month
```

This `JAR.lifetime` parameter is synced with `JAR.expire`, when you update the one parameter, the other is recalculated accordingly:

```php
// JAR.lifetime = 2 hours
Array
(
    [expire] => 1514920993
    [lifetime] => 7200
)

// JAR.lifetime = 3 days
Array
(
    [expire] => 1515172993
    [lifetime] => 259200
)

// JAR.lifetime = 1 month
Array
(
    [expire] => 1517592193
    [lifetime] => 2678400
)
```

### DIFF
--- a/base.php
+++ b/base.php
@@ -395,10 +395,11 @@ final class Base extends Prefab implements ArrayAccess {
 		$ref=$val;
 		if (preg_match('/^JAR\b/',$key)) {
 			if ($key=='JAR.lifetime')
-				$this->set('JAR.expire',is_int($val)?$time+$val:strtotime($val));
+				$this->set('JAR.expire',$val==0?0:
+					(is_int($val)?$time+$val:strtotime($val)));
 			else {
 				if ($key=='JAR.expire')
-					$this->hive['JAR']['lifetime']=($val-$time);
+					$this->hive['JAR']['lifetime']=max(0,$val-$time);
 				$jar=$this->unserialize($this->serialize($this->hive['JAR']));
 				unset($jar['expire']);
 				if (!headers_sent() && session_status()!=PHP_SESSION_ACTIVE)


xfra35:
There's something I don't understand here: why should `JAR.expire` be set to a timestamp since it expresses the ttl of session cookies and the default ttl of other cookies?

1) when we call `$f3->set('JAR.expire',$val)`, we pass a ttl to `session_set_cookie_params`

2) when we call `f3->set('COOKIE.foo','bar')`, we pass a timestamp to `setcookie`, which is basically the current timestamp + the default cookie ttl (coming from `JAR.expire`)

3) when we call `f3->set('COOKIE.foo','bar',$ttl)`, we pass a timestamp to `setcookie`, which is the current timestamp + `$ttl`. **`JAR.expire` is not used here**

Couldn't we handle all those cases with `JAR.expire` being a ttl?

ikkez:
I was just thinking because of backwards compatibility for **JAR.expire**... though it's probably not a big deal.

I think `JAR` originates from [setcookie()](http://php.net/manual/de/function.setcookie.php) because its method signature follows the JAR properties:

```php
setcookie ( $name, $value, $expire, $path, $domain, $secure, $httponly )
```
were `$expire` must be a unit timestamp. In [`session_set_cookie_params()`](http://php.net/manual/de/function.session-set-cookie-params.php) the method signature is similar, however the property is called `$lifetime` and is just the TTL for the session cookie. Actually I'm thinking about handling both values separately but that wouldn't solve #178
But yes, probably some things could be simplified if JAR.expire becomes a TTL int only, ....do we need the possibility to easily set a fixed expiration date for cookies/session lifetime? 🤔

xfra35:
> do we need the possibility to easily set a fixed expiration date for cookies/session lifetime?

Are you talking about this special case:

```php
$f3->set('JAR.expire',strtotime('2018-01-01'));
$f3->set('COOKIE.foo','bar');
```
?

Is there any practical use case for it?

ikkez:
maybe for unit tests? Well but I think it's probably a no brainer... I'll modify the PR to reflect JAR.expire being a ttl, if you think that it would fit better.

xfra35:
I was just wondering if there was any practical use case for setting a fixed non-session cookies expiration timestamp. I don't see any, and even if there was, we could handle that case with a lifetime properly computed.

Anyway we can't break BC, so let's merge this PR, with both `JAR.lifetime` and `JAR.expire` and when time for releasing v4 will come, we'll be free to remove `JAR.expire` and simplify the codebase.

Just a small thing before merging. I don't think you've handled the special value `0`:
```php
$f3->set('JAR.expire',0); // JAR.lifetime should also be set to 0
$f3->set('JAR.lifetime',0); // JAR.expire should also be set to 0
```

ikkez:
I agree to keep BC here.. and it's easy to adjust this for v4, so I'll label this for later and merge the patch for now.

PR #243, opened 2018-02-02
Ensure valid route alias names

xfra35:
I think that the regex in `$f3->route()` is a bit too permissive for aliases. It allows nearly all characters, in particular dots, while it should stick to the `\w` range.

What drew my attention is an issue for my f3-access plugin where the user has defined aliases with dots: https://github.com/xfra35/f3-access/issues/4

Additionally, there's that `?` after `@` which seems superfluous to me. But I could be missing something.

### DIFF
--- a/base.php
+++ b/base.php
@@ -1363,7 +1363,7 @@ final class Base extends Prefab implements ArrayAccess {
 				$this->route($item,$handler,$ttl,$kbps);
 			return;
 		}
-		preg_match('/([\|\w]+)\h+(?:(?:@?(.+?)\h*:\h*)?(@(\w+)|[^\h]+))'.
+		preg_match('/([\|\w]+)\h+(?:(?:@(\w+)\h*:\h*)?(@(\w+)|[^\h]+))'.
 			'(?:\h+\[('.implode('|',$types).')\])?/u',$pattern,$parts);
 		if (isset($parts[2]) && $parts[2])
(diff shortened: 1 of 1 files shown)


ikkez:
We had `\w+` before, and the change to `.+?` was introduced not that far ago and it seems to be made on purpose, see this blame: https://github.com/bcosca/fatfree-core/commit/a27f2f0ef1755f59df949a08030b43f83af0d8b3
Perhaps it was also regarding some unicode support 🤔

xfra35:
You're right, thanks.

@bcosca what was the reason of the tweak in `route()` regex in commit https://github.com/bcosca/fatfree-core/commit/a27f2f0ef1755f59df949a08030b43f83af0d8b3?

Was it to allow unicode characters in route aliases? If yes, we should forbid dots and make the regex consistent in [mock()](https://github.com/bcosca/fatfree-core/blob/10f97fb756b42d82ccdc02ad8054a1af8b589c83/base.php#L1297) as well as in the second part of the [route() regex](https://github.com/bcosca/fatfree-core/blob/10f97fb756b42d82ccdc02ad8054a1af8b589c83/base.php#L1366).

bcosca:
You're absolutely right. The intent of that commit was to allow for unicode characters.

ikkez:
fixed in https://github.com/bcosca/fatfree-core/commit/5164e8c33aa823c74f11d67fb53ea9dfecf079b6

PR #247, opened 2018-03-20
Fix URL parse with PHP < 5.4.7

korukugashi:
Bug : routes are "404 not found //example.com/"

We must provide the protocole in the URL scheme to be compatible with versions of PHP < 5.4.7 and > 5.4

See : http://php.net/manual/en/function.parse-url.php
PHP 5.4.7 - Fixed host recognition when scheme is omitted and a leading component separator is present


xfra35:
Nice spot @korukugashi thanks.

@ikkez what's the PHP requirement now? `composer.json` says `5.4` but I thought it was a bit higher than that.

korukugashi:
I agree that's an outdated version. I cannot update the server for specific constraints.

If you do not take into account this compatibility, which I understand perfectly, you should increase the version number supported by the framework to 5.4.7 and not 5.4.

PR #252, opened 2018-05-24
Method uppercase to avoid route failure

Glandos:
My use case is with simpleid/simpleid project. When trying to log in with OpenID 2 on http://osm.org the 3rd party send a long POST with a `_method=post` in it. I totally don't know if it comes from OpenID standard or not. Nonetheless, this make the authentication failed, because of this parsing.

Is there any rationale behind considering this parameter as the real method?


xfra35:
Thanks

xfra35:
Both `POST._method` and the `X-HTTP-Method-Override` header allow to override the request method. This can come in handy when requesting a REST API from a simple HTML form (i.e non-ajax).

PR #253, opened 2018-08-12
Fixed error description when (PSR-11) `CONTAINER` fails to resolve a class

Rayne:
This PR fixes the class resolving error description, e.g.

```
Invalid class 'ExampleController'
```

instead of

```
Invalid class League\Container\Container::__set_state([])
```

when working with a PSR-11 dependency injection container.


Rayne:
Ping @nanawel

nanawel:
Indeed! :+1:

PR #257, opened 2018-09-15
Fix OPTIONS method for REST requests, now returning only callable methods on handler

nanawel:
I'm proposing this PR to fix what seems to be a bug in the REST method mapping process.

With the given example in the [documentation](https://fatfreeframework.com/3.6/routing-engine#ReST:RepresentationalStateTransfer):
```php
class Item {
    function get() {}
    function post() {}
    function put() {}
    function delete() {}
}

$f3=require('lib/base.php');
$f3->map('/cart/@item','Item');
$f3->run();
```

we have the following behavior:

- calling GET on `/cart` will result in HTTP status 200 with empty body :heavy_check_mark: 
- calling PATCH on `/cart` will result in HTTP status 405 with empty body and the `Allow` header set to `POST,PUT,DELETE,OPTIONS` :heavy_check_mark: 
- calling OPTIONS `/cart` will result in HTTP status 200 with empty body and the `Allow` header set to `GET,HEAD,POST,PUT,PATCH,DELETE,CONNECT,OPTIONS` while obviously not all these methods are supported by the handler here :x:

That's because the `\Base::map()` method creates routes for all `\Base::VERBS` when called, while it should only do for methods that are _callable_ on the handler.

With my PR, calling OPTIONS `/cart` will result in HTTP status 200 with empty body and the `Allow` header set to `POST,PUT,DELETE,OPTIONS` as it should.

Be aware that I prefered using `is_callable()` instead of `method_exists()` because one may want to define a `__call()` method on the handler. In that case, the PR is useless as all methods will be detected as callable because of the catch-all `__call` method.

### DIFF
--- a/base.php
+++ b/base.php
@@ -1464,10 +1464,11 @@ final class Base extends Prefab implements ArrayAccess {
 			return;
 		}
 		foreach (explode('|',self::VERBS) as $method)
-			$this->route($method.' '.$url,is_string($class)?
-				$class.'->'.$this->hive['PREMAP'].strtolower($method):
-				[$class,$this->hive['PREMAP'].strtolower($method)],
-				$ttl,$kbps);
+			if (is_callable([$class,$this->hive['PREMAP'].strtolower($method)]))
+				$this->route($method.' '.$url,is_string($class)?
+					$class.'->'.$this->hive['PREMAP'].strtolower($method):
+					[$class,$this->hive['PREMAP'].strtolower($method)],
+					$ttl,$kbps);
 	}
 
 	/**


Rayne:
Good catch! Isn't this change breaking the (basic) `callable` support of `Base->map()`?

```php
$f3->map('/cart/@item', function($f3, $args){ /* … */ });
```

KOTRET:
@Rayne thanks for pointing this out.
There should be a test for that to ensure correct behaviour

bcosca:
I have some misgivings about this pull request as I for one use `__call` as a catch-all for fine-grain control.

nanawel:
@bcosca As I said, this PR will only have an effect if used with controllers that **don't** define `__call()` method. Otherwise, behavior will be left unchanged (because `is_callable()` will return `true` for all verbs).

Was that your worry?

bcosca:
A better solution is to fix the `run()` code so it checks the route handling class if an `options()` method exists. That way, a user can override the default behavior.

bcosca:
@nanawel Please look at commits fatfree#086a8b and fatfree/core#3260f39.

nanawel:
> A better solution is to fix the `run()` code so it checks the route handling class if an `options()` method exists. That way, a user can override the default behavior.

That's much better indeed. Actually that was how I supposed it was implemented at first, but then I saw that the `options()` method was not called and when I traced back how it works I found this "bug" I tried to fix the simplest way I could.

I don't have access to commit  bcosca/fatfree@3260f39 btw (404).

bcosca:
Just grab the latest snapshot and you should get the mentioned fix. Closing this now.

bcosca:
There now. Sorry about that.

nanawel:
@bcosca There's still a problem now with the `Access-Control-Allow-Methods` header which is part of the CORS implementation. Even if the `options()` method sets the correct value for this header it gets overwritten afterward.

The fix is the same I guess, with `preg_grep()` just before the block.

ikkez:
actually when you plan to implement cors headers yourself in the options handler, you dont need to enable the CORS feature in the framewrok, and dont need to patch anything more here.

nanawel:
@ikkez That's right, but the framework already provides some elements, like the handling of the `Origin` header, so you just have to set `CORS.origin=*` in you config file for example to enable it.

It could just be a bit better if you could just implement the allowed methods header part, just like `Allow` now.

nanawel:
Should I propose a PR for this too? Or is it settled that CORS have to be implemented by the developer because what is proposed by the framework is not sufficient and won't be?

ikkez:
can you please check my latest commit and if that solves the issue?!

nanawel:
From what I see yes it should. This seems similar to the fix I implemented for my case.

PR #264, opened 2018-11-16
Update Cache

extralam:
Add 2 helper function to using the cache system.
Update Redis cache using delete and setex.


ikkez:
Sorry, the Redis function names are not actually needed to change, and your helper methods are similar to what Base->set and ->get are already offering. To keep it fat free I'll close this here, but thanks for your effort.

PR #271, opened 2019-02-02
Add json support for translations

jtojnar:
This will allow us to share translations with JavaScript, as well as better interoperability with things like Weblate or CrowdIn.

### DIFF
--- a/base.php
+++ b/base.php
@@ -1083,6 +1083,9 @@ final class Base extends Prefab implements ArrayAccess {
 					is_file($file=$base.'.php')) &&
 					is_array($dict=require($file)))
 					$lex+=$dict;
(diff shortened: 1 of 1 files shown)


eazuka:
Thanks to @jtojnar for this and to you @ikkez for accepting and merging this PR.

So this mean translations can be in json format and having .json file extension?

Rayne:
@eazuka That's correct. You have to setup `LOCALES` as usual and supply `*.json` files. If it doesn't work for you then it could be possible that you are using the latest stable release (3.6.5) instead of the development version.

eazuka:
thanks @Rayne, i'm using dev version at the moment. i'm aware the json support isn't available yet on the stable (3.6.5) release.

PR #285, opened 2019-09-26
Throw error on mapper update/erase if the table has no primary key

xfra35:
I just had a bad experience today with the `DB\SQL\Mapper->save()` executing an `UPDATE` on the whole table instead of a single row. I realized that the table had a unique key but no primary key... and the framework did nothing to prevent the massive undesired `UPDATE`.

Hence this proposal to throw an error on `update/erase()` if no PK can be found.

### DIFF
--- a/db/sql/mapper.php
+++ b/db/sql/mapper.php
@@ -25,6 +25,11 @@ namespace DB\SQL;
 //! SQL data mapper
 class Mapper extends \DB\Cursor {
 
+	//@{ Error messages
+	const
+		E_PKey='Table %s does not have a primary key';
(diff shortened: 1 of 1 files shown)


KOTRET:
i welcome this pr, although i can imagine of config-tables with only one row in it, so that this breaks compatibility.

ikkez:
I'm actually for this change, as I also had a similar adventure once before. If there are edge-cases for 1-row tables, we'll find a solution for it. I could imagine to put this behaviour into a mass-update method, so that you can batch-update all or some records at once.

PR #290, opened 2019-12-26
Error inserting data

tcordel:
+----------------------+--------------+------+-----+---------+-------+
| Field                | Type         | Null | Key | Default | Extra |
+----------------------+--------------+------+-----+---------+-------+
| bookingStayId        | varchar(20)  | NO   | PRI | NULL    |       |
| bookingRoomNum       | int(8)       | NO   | PRI | NULL    |       |
| multiRoomNumber      | int(8)       | NO   | PRI | 0       |       |

Primary key multiRoomNumber with default value as 0 is seen as empty --> taken as primary key
This leads to an outofmemory.

### DIFF
--- a/db/sql/mapper.php
+++ b/db/sql/mapper.php
@@ -447,7 +447,7 @@ class Mapper extends \DB\Cursor {
 			if ($field['pkey']) {
 				$field['previous']=$field['value'];
 				if (!$inc && $field['pdo_type']==\PDO::PARAM_INT &&
-					empty($field['value']) && !$field['nullable'])
+					is_null($field['value']) && !$field['nullable'])
 					$inc=$key;
 				$filter.=($filter?' AND ':'').$this->db->quotekey($key).'=?';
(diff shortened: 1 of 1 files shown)


ikkez:
Hi again @tcordel 
Can you please describe once more what the real issue was? I ask because other problems occur now.. Do i see correctly that you're using a composite primary key in this table based on 3 column?! and on an insert, the default 0 causes the mapper to try to reload the inserted record based on that field, which you actually wanted to forbid, yes? I'm currently thinking about if proper handle for multi-column pkeys in the reload procedure would be a better fix 🤔

tcordel:
Hi Christian,

Yes, you catch the(my) defect.
Maybe you're right about doing this into the handler. I can't figure out
why this update may lead to new issues if you have any exemples i may think
about it too,

Regards,

TC

Le mar. 4 févr. 2020 à 14:14, Christian Knuth <notifications@github.com> a
écrit :

PR #307, opened 2020-11-03
Performance optimization with update to pre-increment/decrement

Weshley:
Updated all occurrences of inc/decremented values to utilize pre- rather than post- operators. Post-inc/decremented operations create a copy of the variable and therefore yield overhead to the operation. 

Sources:
https://stackoverflow.com/questions/1756015/whats-the-difference-between-i-and-i-in-php

https://www.php.net/manual/en/language.operators.increment.php

### DIFF
--- a/audit.php
+++ b/audit.php
@@ -145,7 +145,7 @@ class Audit extends Prefab {
 			return FALSE;
 		$id=strrev($id);
 		$sum=0;
-		for ($i=0,$l=strlen($id);$i<$l;$i++)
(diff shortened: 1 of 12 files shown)


n0nag0n:
It's a good simple micro optimization I can't imagine would hurt things too much. Probably won't notice the performance improvement, but on larger projects that use these iterators often, you might notice a slight bump in performance.

ikkez:
Many drops make a lake. A performance lake :D seems like it's faster yes, only on the big picture though, but valid.

PR #309, opened 2020-11-13
Added primary key length for mysql

n0nag0n:
This table fails to create if your default charset is utf8mb4 for your database because the key length of 255 is too long. This should fix it.

### DIFF
--- a/db/sql/session.php
+++ b/db/sql/session.php
@@ -187,7 +187,7 @@ class Session extends Mapper {
 					$tab.$db->quotekey('ip').' VARCHAR(45),'.$eol.
 					$tab.$db->quotekey('agent').' VARCHAR(300),'.$eol.
 					$tab.$db->quotekey('stamp').' INTEGER,'.$eol.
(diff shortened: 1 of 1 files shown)


ikkez:
Depending on [session.hash-function](https://www.php.net/manual/en/session.configuration.php#ini.session.hash-function) in php.ini, the session ID might be longer than just 20. With utf8mb4 you can only index 191 chars in certain version, so I guess it'll be better to cap this so 128 here.

ikkez:
Just tried it myself with `utf8mb4_unicode_ci` and the current version and there's no problem to create the table, column session_id (255) and the default index. MySQL 5.7.26  / PHP7.4

ikkez:
Okay this only seems to be an issue for older databases / tables in MYISAM format. Since that is not the default anymore since (mysql 5.5) we can probably ignore this.

n0nag0n:
Hmm I'll have to try it again on a new project to see if it's still an issue. I had to do this manually for like 3 projects because of the length issue about .... 4-5 months ago

PR #312, opened 2020-12-10
DB\SQL: Fix handling of PDO prepare() errors

Alanaktion:
This changes the DB\SQL class to use the PDO instance's errorInfo() instead of attempting to call a nonexistent function. I believe this was what was originally intended with this code.


ikkez:
Yes I think so too. Thanks

PR #324, opened 2020-12-17
Allow mock() to set the _POST arguments for PUT and PATCH mehtods

Frolki1-Dev:
I had a problem in my tests, that PUT and PATCH ignored the arguments that i passed.

With this pull request I fix this, that also on a PUT or PATCH request will pass the arguments in the _POST.


ikkez:
But isn't payload from PUT (and PATCH?) usually found in BODY var instead of POST?

Frolki1-Dev:
My fault, sorry. Normal you can fetch the inputs in PHP with `php://input`. But in the mock it set in the hive with the key `BODY`.

Now we can decide between:
* For mock we allow to set all body information into $_POST or $_REQUEST
* Grab in the testing enviroment the body data from the hive and merge it with `php://input`

Here is my code in the Controller: (It's in the __construct method)
image

PR #329, opened 2021-07-09
Fix CACHE directory duplicate

jakofff:
When the global CACHE has the value of the cache directory, it duplicates the creation of the directory, creates that of the variable and that of default

### DIFF
--- a/base.php
+++ b/base.php
@@ -2801,7 +2801,7 @@ class Cache extends Prefab {
 						$this->ref=new Memcached();
 					$this->ref->addServer($host,$port);
 				}
-			if (empty($this->ref) && !preg_match('/^folder\h*=/',$dsn))
+			if (empty($this->ref) && $dsn===TRUE)
 				$dsn=($grep=preg_grep('/^(apc|wincache|xcache)/',
 					array_map('strtolower',get_loaded_extensions())))?
 						// Auto-detect


n0nag0n:
That is a very small change, but I have a funny feeling it's going to break something else...

jakofff:
I explain the fix ....

Refers to duplication of directory creation for CACHE.

In the documentation it indicates that CACHE can have 3 possible values, TRUE, FALSE and "folder =", by default it is FALSE.

When the value is FALSE, no directory is created.

When the value is TRUE there is no problem since the default directory is created in "tmp/cache" relative to the path where the project is.

But the issue is when CACHE has the type value "folder=/something/ cache" what it does is that the directory that has "folder" is created but the default directory "tmp/cache" is also created.

The fix what it does is that in that if block the default directory is created, therefore the only thing missing is only to validate that when CACHE is TRUE it creates the default directory.

xfra35:
Hi @jakofff,

If there is a bug, I don't think it is where you think.

Indeed, there's only one `mkdir` in this function, so if you're ending up with 2 directories created, that means the function was called twice with 2 different parameters.

How exactly do you instantiate and configure the Cache class?

base.php
```diff
@@ -2801,7 +2801,7 @@ function load($dsn,$seed=NULL) {
 						$this->ref=new Memcached();
 					$this->ref->addServer($host,$port);
 				}
-			if (empty($this->ref) && !preg_match('/^folder\h*=/',$dsn))
+			if (empty($this->ref) && $dsn===TRUE)
```
xfra35, line 2804: `$dsn` would never equal `TRUE`, because of the `$dsn=trim($dsn)` call above.

jakofff:
In the function, dsn takes the value that has cache, by default it is a FALSE it does not matter if a trim is performed, at the end it can be a FALSE, TRUE, or a Folder = ""

The issue is that in the last two blocks they are if, they are not if else, therefore it will try to enter both, in the first if there is no module loaded the result will be the path of the default directory "('folder =' . $ fw-> TEMP.'cache / ') ", in the second if create the directory based on dsn and the directory part.

The fix solves just this, that if CACHE has the value of true, it creates only the default directory, if it has FALSE it does nothing, and if it has "folder" it creates only the directory indicated.

xfra35:
> it does not matter if a trim is performed, at the end it can be a FALSE, TRUE, or a Folder = ""

Yes it does: `trim(TRUE)` returns `"1"`, not `TRUE`.

> The issue is that in the last two blocks they are if, they are not if else, therefore it will try to enter both, in the first if there is no module loaded the result will be the path of the default directory "('folder =' . $ fw-> TEMP.'cache / ') ", in the second if create the directory based on dsn and the directory part.

That's correct, and this is not an issue:
- in the first block, it autodetects cache extensions (APC/Wincache/Xcache) and fallbacks to folder cache. In the latter case, `$dsn` is replaced with `folder=tmp/cache/` or similar.
- in the second block, it checks if the cache folder exists (no matter if it's the default folder or not) and creates it otherwise.

This is how it is supposed to be, and I don't see anything wrong in it.

Please share your code.

jakofff:
> Yes it does: trim(TRUE) returns "1", not TRUE.

exactly why in the fix the comparison is to TRUE not to, because when CACHE = TRUE the value passes as TRUE. That's what I meant when I mentioned that the trim did not matter, because the value that CACHE can have is TRUE

> That's correct, and this is not an issue:

in the first block, it autodetects cache extensions (APC/Wincache/Xcache) and fallbacks to folder cache. In the latter case, $dsn is replaced with folder=tmp/cache/ or similar.
in the second block, it checks if the cache folder exists (no matter if it's the default folder or not) and creates it otherwise.

What you say is also correct, only that when it enters the first if, if the value is FOLDER, it will always take the default directory temp / cache.

That is why when it reaches the second, it creates it.

> To do the test it is very easy in a config or ini put different values to cache, those that are in the documentation

CACHE = FALSE -> default value and does not use cache therefore does not create cache directory

CACHE = TRUE -> uses file-based cache and with the default directory tmp / cache relative to the application directory.

CACHE = "folder: / dir / cache" -> create the directory to save the sessions.

ikkez:
I'm closing this because I cannot reproduce the issue either. It's perhaps rather an issue in your application, where you have set CACHE to true somewhere before, because this is the only situation where the default and a custom folder is created:

```php
$f3->set('CACHE', TRUE);
$f3->set('CACHE', 'folder=tmp/foo/');
```

PR #332, opened 2021-12-02
Fix PHP 8.1.0 compatibility

slicke:
Address errors preventing F3 from running on PHP 8.1

### DIFF
--- a/base.php
+++ b/base.php
@@ -713,7 +713,7 @@ final class Base extends Prefab implements ArrayAccess {
 	**/
 	function split($str,$noempty=TRUE) {
 		return array_map('trim',
-			preg_split('/[,;|]/',$str,0,$noempty?PREG_SPLIT_NO_EMPTY:0));
+			preg_split('/[,;|]/',$str??'',0,$noempty?PREG_SPLIT_NO_EMPTY:0));
 	}
 
 	/**
--- a/base.php
+++ b/base.php
@@ -2234,8 +2234,8 @@ final class Base extends Prefab implements ArrayAccess {
 	*	@return mixed
 	*	@param $key string
 	**/
-        #[\ReturnTypeWillChange]
-        function offsetexists($key) {
+    #[\ReturnTypeWillChange]
+    function offsetexists($key) {
 		return $this->exists($key);
 	}
 
@@ -2244,8 +2244,8 @@ final class Base extends Prefab implements ArrayAccess {
 	*	@return mixed
 	*	@param $key string
 	*	@param $val mixed
-        **/
-        #[\ReturnTypeWillChange]
+    **/
+    #[\ReturnTypeWillChange]
 	function offsetset($key,$val) {
 		return $this->set($key,$val);
 	}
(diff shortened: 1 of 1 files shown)


base.php
```diff
@@ -713,7 +713,7 @@ function fixslashes($str) {
 	**/
 	function split($str,$noempty=TRUE) {
 		return array_map('trim',
-			preg_split('/[,;|]/',$str,0,$noempty?PREG_SPLIT_NO_EMPTY:0));
+			preg_split('/[,;|]/',$str??'',0,$noempty?PREG_SPLIT_NO_EMPTY:0));
```
joanhey, line 716: This change need PHP7 and break the requirements of php >= 5.4:
 https://fatfreeframework.com/3.7/system-requirements

Using the elvis operator (php5.3) will work without need a higher version.
`$str ?: ''`

slicke: Good catch! :) Fixed it!

base.php
```diff
@@ -2234,7 +2234,8 @@ function unload($cwd) {
 	*	@return mixed
 	*	@param $key string
 	**/
-	function offsetexists($key) {
+        #[\ReturnTypeWillChange]
+        function offsetexists($key) {
```
jtojnar, line 2238: This changes the indentation from tabs to spaces. Same issue on most other changed lines.

base.php
```diff
@@ -2234,7 +2234,8 @@ function unload($cwd) {
 	*	@return mixed
 	*	@param $key string
 	**/
-	function offsetexists($key) {
+        #[\ReturnTypeWillChange]
```
jtojnar, line 2237: Maybe we should just change the method to do `!== false` check so that boolean is always returned. This would be technically a BC break but hopefully no-one is using `ArrayAccess::offsetExists` to get `ttl`. And if they do, they can trivially replace it with `exists`.

base.php
dduers: Thank you!
One thing - the elvis operator is not null safe.
Maybe it's no matter in the context, but otherwise this will mitigate:
```
isset($str) ? $str : ''
```

base.php
jtojnar: That will not actually help since PHP shows the warning based on return type hint, even for child class methods that have no type hints: https://php.watch/versions/8.1/internal-method-return-types#no-return-type

dduers: Seems there is no way around breaking with php5.4
Or include some circumstances.

maietta: > 

Composer is a great utility because people can lock in versions of dependencies their application may require via the composer.json file. We shouldn't be worried so much trying to support ancient versions of PHP.

base.php
slicke: Sorry! I think I fixed it now :)

jtojnar: There are still several instances. You can use [Refined GitHub](https://github.com/refined-github/refined-github) extension to make the whitespace visible.

base.php
jtojnar: The ternary operator works with any truthy/falsy value. The `isset` is only needed when you want to e.g. fetch an array index that might not exist.

base.php
jtojnar: I fixed the indentation as well as added descriptions in the commit messages and few more fixes in #333.

ikkez:
thanks for the patience. It's all fixed now.

PR #361, opened 2023-04-14
Adds support for geometry fields in SQL mappers

Lawrence72:
Currently its not possible to insert/update geometry fields with a SQL mapper. 

This adds support for this

How to use: 

```
$this->User_Mapper->position =  'POINT (30 10)';
$this->User_Mapper->save();
```

This also works with other geometry types: 

`$this->User_Mapper->position =  'LINESTRING(2 1, 6 6)';`

`$this->User_Mapper->position =  'POLYGON((0 5, 2 5, 2 7, 0 7, 0 5))';`

Retrieving geometry information can be done using a virtual field

```
$this->User_Mapper->distance = "ST_Distance(POINT(2,1), position)";
$this->User_Mapper->load(['id = ?', 1]);
echo $this->User_Mapper->distance;
```

### DIFF
--- a/db/sql/mapper.php
+++ b/db/sql/mapper.php
@@ -464,7 +464,7 @@ class Mapper extends \DB\Cursor {
 			}
 			if ($field['changed'] && $key!=$inc) {
 				$fields.=($actr?',':'').$this->db->quotekey($key);
-				$values.=($actr?',':'').'?';
+				$values.=($actr?',':'').($field['type'] === 'geometry' ? 'ST_GeomFromText(?)' : '?');
 				$args[$actr+1]=[$field['value'],$field['pdo_type']];
 				++$actr;
 				$ckeys[]=$key;
@@ -527,7 +527,7 @@ class Mapper extends \DB\Cursor {
 			return $this;
 		foreach ($this->fields as $key=>$field)
 			if ($field['changed']) {
-				$pairs.=($pairs?',':'').$this->db->quotekey($key).'=?';
+				$pairs.=($pairs?',':'').$this->db->quotekey($key).($field['type'] === 'geometry' ? '=ST_GeomFromText(?)' : '=?');
 				$args[++$ctr]=[$field['value'],$field['pdo_type']];
 			}
 		if ($pairs) {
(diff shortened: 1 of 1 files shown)


ikkez:
@Lawrence72  How is the field type `geometry` detected, or did you set that manually somewhere?

Lawrence72:
@ikkez In the mapper class, as part of the construct it grabs the table schema, and part of that is the field types (F3 Mapper already does this), this is all stored in the $this->fields array.  So I am checking the info stored in there to see if the field is set to type geometry and behaving a little differently if it is. Its not a user set variable or anything the user can alter in their mapper call.

ikkez:
Hi again.. sorry for the delay. Does this work for all sql engines or have you only tried MySQL/MariaDB ?
I'll definitely have a look at this again.

Lawrence72:
Ive only tested it on MYSQL/Maria. 

Here is a gist of the test I ran to work out what the issue was:
https://gist.github.com/Lawrence72/0fada42e673abd41600190b4035dfad3

The SQL queries work fine, the select from a mapper works as long as its added as a virtual field 

The insert does not work. What I found was that the mapper didnt handle SQL functions in the mapper field value which is required in order to add a geo spacial.  ( ST_GeomFromText('POINT(4 4)') )

So the solution I was trying to do was to take out the need for the ST_GeomFromText by detecting that the field  was of type geometry and wrap it in a ST_GeomFromText($value) so the mapper could be set to $mapper->field = 'POINT(4 4)' and have it insert correctly. 

The alternative that I can think of for this would be a more general fix where the mapper allows SQL functions in the mapper field value. 

Though when I was looking at it I thought that might cause problems.

Another idea I had was to allow something similar to virtual fields in an insert. That would make it so SQL functions could be added but under more controlled conditions than trying to detect when the value includes a SQL function and when it doesn't. 

I hope this makes sense, on what I was trying to do with it.

KOTRET:
the function to call depends on the dbms. For SQL Server it's like `geometry::STGeomFromText('<geometry_tagged_text>', <SRID >)`, so this maybe should not be hardcoded at this point

PR #369, opened 2023-12-06
Use the same string replacement for keys in cache getter as setter

BrewCurious:
This is a bit of a band-aid approach to deal with a specific use case for myself. What should eventually happen is a reevaluation of all `is_file`, `write` and `read` calls in `base.php` to have the filename strings always get passed through the `fixslashes` method before checking the filesystem to ensure consistent and filesystem-safe keys.

### DIFF
--- a/base.php
+++ b/base.php
@@ -2648,7 +2648,8 @@ class Cache extends Prefab {
 				$raw=xcache_get($ndx);
 				break;
 			case 'folder':
-				$raw=$fw->read($parts[1].$ndx);
+				$raw=$fw->read($parts[1]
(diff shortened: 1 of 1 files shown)


ikkez:
thanks.. I'll keep this in mind to review all file path handling with fixslashes.. I'll merge this nevertheless for now ;)

BrewCurious:
@ikkez Thanks for the merge. Could you let me know what the typical process/timeline looks like for updates made to `master` here getting updated via the link in the `bcosca/fatfree-core` repo? That's the package that I'm connecting to in my project's composer's config right now.

PR #377, opened 2024-05-03
When querying the fields for a table, the schema is missing

mhdev90:
We have more than one schema in a PGSQL database. This causes problems if tables with the same name exist in different schemas that have different fields.
The problem is that the schema is not used when determining the fields.
I have now added the schema to the WHERE clause. Unfortunately, I can only test it with PGSQL at the moment.


ikkez:
finally tested with sql server as well and it works fine.

PR #379, opened 2024-06-01
Adding optional support for trailing slash clean in PATH

rippak:
Moved one old project to F3 where trailing slashes in routes are obligatory

### DIFF
--- a/base.php
+++ b/base.php
@@ -1728,7 +1728,8 @@ final class Base extends Prefab implements ArrayAccess {
 			if (!$route)
 				continue;
 			if (isset($route[$this->hive['VERB']]) && !$preflight) {
-				if ($this->hive['VERB']=='GET' &&
+				if ($this->hive['PATH_SLASH_CLEAN']===TRUE &&
+                    $this->hive['VERB']=='GET' &&
 					preg_match('/.+\/$/',$this->hive['PATH']))
 					$this->reroute(substr($this->hive['PATH'],0,-1).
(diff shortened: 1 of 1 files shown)


base.php
```diff
@@ -2530,6 +2531,7 @@ function($level,$text,$file,$line) {
 			'PACKAGE'=>self::PACKAGE,
 			'PARAMS'=>[],
 			'PATH'=>$path,
+            'PATH_SLASH_CLEAN'=>TRUE,
```
Rayne, line 2534: When this gets accepted, make all indentations consistent and don't forget to update the documentation. :-)

rippak: Thanks, I'll check it. Phpstorm shows normal indents. 
Where should I update the docs?

Rayne: The docs are [here](https://github.com/f3-factory/F3com-data), but I would wait until this change got accepted.

rippak: Of course, just asking

ikkez:
Would you mind if we rename the option var to something like... `REROUTE_TRAILING_SLASH` maybe? 🤔 
I'm wondering if there's more to do here. Like generating links to aliases? Switching to trailing slashes and also reroute non-trailing slash urls to those with slash, when the new option is on. Having both options enabled might lead to duplicate content warning .. just keep that in mind.

rippak:
Naming is not a problem at all.

ikkez:
Renamed to `REROUTE_TRAILING_SLASH` and committed.

PR #384, opened 2024-12-28
Property for system variables

Vladzimir:
Suggestion all system variables in IDE
`\Base::instance()->IP`


base.php
```diff
@@ -38,7 +38,82 @@ static function instance() {
...
 }
-
+/**
+ * @property string ALIAS
+ * @property array ALIASES
+ * @property string|array AUTOLOAD
+ * @property bool|string CACHE
+ * @property bool CASELESS
+ * @property callable|\Prefab|\Psr\Container\ContainerInterface CONTAINER
+ * @property string
```
ikkez, line 48:
```suggestion
```

PR #385, opened 2024-12-30
Use handler as array ['ClassName::class', 'methodName'] for route

Vladzimir:
Use handler in route as ['ClassName::class', 'methodName'] for hints and clickability in IDE


ikkez:
Already implemented in v4 in a slightly different way. Closing this one.

PR #398, opened 2025-06-21
A bug fix

myaaghubi:


geniuswebtools:
Providing the trailing slash is intended, it's not a bug.  It's also mentioned in the documentation:
[N.B.1: Like every F3 directory name, $dir must end with a trailing slash!
](https://fatfreeframework.com/3.8/jig)

myaaghubi:
Let's have the mapper
```php
$db=new DB\Jig('db/data/',DB\Jig::FORMAT_JSON);
$user=new DB\Jig\Mapper($db,'users');
$user=new DB\Jig\Mapper($db,'categories');
...
```
For a path like `test/dir/db`, it will make dbs like `dbusers`, `dbcategories`, which is not the responsibility of the mapper to fix the issue
The point is whether these small things should be fixed in (it is a bug) or out of the framework

PR #406, opened 2026-01-12
Middleware Support [v4]

ikkez:
Hey folks. This is a new feature for the version 4. 
I've started with v4 years ago.. Since then, my job has changed, priorities, and also the php language itself. Since then there were ups and downs with investigating on how to proceed and renovate the framework core for a more modern age.
Well, time has come. The main refactoring on the Base is done and now the fun part can begin again. More on updates on a later point, but this one introduces a real middleware aware router approach at the core level.

A middleware can be used to send additional headers, do certain authentication checks or manipulate the actual response.
It can be configured similar to routes. It is possible to register them with a pattern: 

```php
$this->f3->middleware('GET|POST|PUT /api/*', function (\Closure $next) {
    // do auth checks here
    return $next();
});
```

When you now open `/api/user/info` the middleware is executed before the actual route handler. Within the middleware you can decide if you want to do something before or after the `$next` handler, which is given as 1st parameter in the closure.

You can also define route tags now and register the middleware with tag as well instead of a pattern:

```php
$this->f3->middleware('auth', function (\Closure $next) {
    // do auth checks here
    return $next();
});

$this->f3->route('GET /api/foo', function () {
    // 
}, tags: ['auth]);
```

You can also define multiple tags or mix it with patterns:

```php
$this->f3->middleware([
    'auth', 
    'api',
    'GET|POST /api/*'
], function (\Closure $next) {
    return $next();
});
```

For config files, the syntax for route tags use a `#` to indicate a tag. There also can be multiple, it goes like this:

```ini
[ROUTES]
GET|POST @alias #tag1 #tag2 /api/user = User->search

[MIDDLEWARES]
tag1 = App\Middleware1
tag2 = App\Middleware2
GET|POST /api/* = App\Middleware3
```

---

Another useful example it the CORS feature for specific paths:

```php
$this->f3->middleware('OPTIONS /api/*', function (\Closure $next, \F3\Base $f3) {
    $f3->CORS['origin'] = 'https://fatfreeframework.com';
    return $next();
});

$this->f3->route('GET /api/user/info', [UserInfo::class, 'get']);
```

When cross side JS does a preflight OPTIONS request now, the proper `Access-Control-*` headers are set, because the middleware enables this feature in the router, but only for the /api/* routes.

---

Since fat-free v4 can use dependency injection and also handles PSR-7 http messages, you can work with these as well in the middleware:

```php
$this->f3->middleware('GET /api/*', function (\Closure $next, ServerRequestInterface $request, ResponseInterface $response) {
    if ($request->getHeader('X-App-Version')[0] < '4.0.1') {
        return $response->withBody(new Stream('Version too old.')); 
    }
    return $next($response);
});
$this->f3->middleware('GET /*', function (\Closure $next, ResponseInterface $response) {
    $response = $next($response);
    return $response->withHeader('X-Foo', 'bar');
});
$this->f3->route('GET /api', function (\F3\Base $f3, ResponseInterface $response) {
    return $response->withBody(new Stream('Hallo World!'));
});
```

Unit tests are provided at the fatfree-dev repo. Thanks for your attention.

### DIFF
--- a/F3/Base.php
+++ b/F3/Base.php
@@ -2185,7 +2185,7 @@ namespace F3 {
                         if ($match['section']) {
                             $sec = $match['section'];
                             if (\preg_match(
-                                    '/^(?!(?:global|config|route|map|redirect)s\b)'.
+                                    '/^(?!(?:global|config|route|middleware|map|redirect)s\b)'.
                                     '(.*?)\s*:/i',
                                     $sec,
                                     $msec,
@@ -2193,7 +2193,7 @@ namespace F3 {
                                 !$this->exists($msec[1]))
                                 $this->set($msec[1], null);
                             \preg_match(
-                                '/^(config|route|map|redirect)s\b/i',
+                                '/^(config|route|middleware|map|redirect)s\b/i',
                                 $sec,
                                 $cmd,
                             );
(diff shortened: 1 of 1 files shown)


Rayne:
Looks pretty useful. I wish I had time to tinker with it …

itsekomagain:
Still the lightest in 2026

PR #416, opened 2026-08-16
fix: SMTP TLS failure check, CRLF injection, and cert verification default

GhaziTriki:
Fixes three SMTP security issues:

- **TLS handshake failure ignored** — `stream_socket_enable_crypto()` return value was unchecked; credentials sent over unencrypted connection on failure. Now aborts.
- **CRLF injection in MAIL FROM / RCPT TO** — Email addresses were not sanitized for `\r\n`, allowing SMTP command injection. Now stripped.
- **No TLS cert verification by default** — SSL context had no `verify_peer`/`verify_peer_name`. Now defaults to TRUE (user context still overrides). Also removed deprecated TLSv1.1.


ikkez:
thank you ❤️

